PR #6, opened 2022-04-16
Minimal misc changes to improve developments

YonLiud:


ArjixWasTaken:
10/10 PR

Relected:
that's actually pretty cool

baiorett:
this is a useless contribution

Lucaskyy:
@baiorett Watch your words next time, thank you.

baiorett:
> @baiorett Please act respectfully, thank you.

i mean, in which way this improves development

Lucaskyy:
Even if you find this contribution "useless", you act respectfully and professionally. Every contribution has value.

baiorett (commented):
-

YonLiud:
> again, i don't see any reason in this small change because the user will not even see that because integrations are not meant to be run but built

it's not to help the end-user, it's to help other developers like you to debug incase of a problem

baiorett:
> > again, i don't see any reason in this small change because the user will not even see that because integrations are not meant to be run but built
> 
> it's not to help the end-user, it's to help other developers like you to debug incase of a problem

the debugging is handled in gradle itself, not in the batch scripts that initialize gradle

lewis1277:
> again, i don't see any reason in this small change because the user will not even see that because integrations are not meant to be run but built

Only thinking of the end-user? tut tut tut

ReduxFlakes:
This should be implemented

Lucaskyy:
Thank you for this PR!

ponces:
The renaming of the gradlew script does not make sense and it's not coherent with the other revanced repos.

> that's actually pretty good, the gradlew.sh file should make it easier for Linux users to recognize the files

Gradle Linux users know that gradlew is the script to handle gradle tasks...

Please revert this part of the PR :(

oSumAtrIX:
I agree to @ponces

ArjixWasTaken:
Objection!

I would think that a file without a file extension is a binary file, so I wouldn't try to execute it.
But since @Lucaskyy renamed it to have the .sh extension it is now obvious that it is a script we can execute through the command line.

As for it not being coherent with the other repos, that's a skill issue.

Lucaskyy:
> Objection!

Hearsay!

Sorry, watched too many Johnny Depp vs Amber Heard videos.

tillnelown:
Why not remove the '.sh' and instead give the file executable permission? This highlights the file and shows everyone which file to execute through ls and so on.

PR #9, opened 2022-05-26
rename gradlew.sh

tillnelown:
To maintain consistency across repos and keep with gradles [default](https://gradle.org/install/), rename gradlew.sh to gradlew.


Lucaskyy:
Thank you!

oSumAtrIX:
As a note, the integrations are currently in an unmaintained state. Partially decompiled code is also subject to be cleaned up.

PR #21, opened 2022-06-25
fix: some more refactoring and cleanup

TheJeterLP:
PR is named fix so a new version gets generated

Needs to be merged together with https://github.com/revanced/revanced-patches/pull/63


OxrxL:
@TheJeterLP Technically you could remove VP9 codec override (with model override too), due to features deprecation.

At least based on what @oSumAtrIX said.

TheJeterLP:
> @TheJeterLP Technically you could remove VP9 codec override (with model override too), due to features deprecation.
> 
> At least based on what @oSumAtrIX said.

@oSumAtrIX can you confirm that this doesn't work anymore?

oSumAtrIX:
I never claimed that.

OxrxL:
> I never claimed that.

Well...maybe was I my mistake.

Also if I remember that from an old private discussion. 🤷‍♂️

PR #24, opened 2022-06-25
fix: home ads

OxrxL:


TheJeterLP:
Did you test this?

OxrxL:
> Did you test this?

Even if I want to, I can't.

I'm not able to see these ads.

TheJeterLP:
> > Did you test this?
> 
> Even if I want to, I can't.
> 
> I'm not able to see these ads.
> 
> This is a test addressed to those who view ad panels on home.

Google home? So that happens only on root? Or how could I try to reproduce this?

PR #46, opened 2022-07-03
'Suggested for you' block

OxrxL:
https://github.com/revanced/revanced-patches/issues/109


oSumAtrIX:
Please add a new setting for that. People might want to block ads but not recommendations.

oSumAtrIX:
Thanks

PR #50, opened 2022-07-04
feat: `swipe-controls`

shadow578:
Adds video swipe controls (aka `FensterV2`) that enable the user to change volume and screen brightness by swiping. 
This implementation **replaces** the old one, since i didn't feel like rewriting _that_ mess...

For accompanying patch see https://github.com/revanced/revanced-patches/pull/115

## Features
- Swipe controls for volume and screen brightness
- Swipe controls are only enabled after a long- press, to avoid changing volume on accident
- Fling-to-mute feature: when swiping down really fast on the right screen side, volume is instantly set to zero
- Haptic feedback for long-press and Fling-to-mute
- Feature enablement controls are included, but stubbed

## TODOs
- [X] Volume Swipe Sensitivity is too high
- [X] Fine tune Swipe Zone sizes

## Screenshots

| UI Overlay | Swipe Zones Highlighted|
| ------- | ------- |
| big-bunny | swipe-zones  |


shadow578:
hi, i added the SettingsEnum calls into the enablement class. Currently still commented, as i just cannot get SettingsEnum to work correctly (it always returns false, even on fresh install)...

the settings `SWIPE_THRESHOLD_INTEGER`, `SWIPE_PADDING_TOP_INTEGER`, and `SWIPE_USE_TABLET_MODE` are no longer applicable

TheJeterLP:
You did SettingsEnum.Settingname.boolean, that's wrong. It's actually a getter method, so use getBoolean()

TheJeterLP:
That's why I didn't code stuff in kotlin yet. Because it's just bad to mix java with kotlin

bogadana:
no, this is the right way to do it in kotlin

oSumAtrIX:
> That's why I didn't code stuff in kotlin yet. Because it's just bad to mix java with kotlin

From the official Kotlin docs:
>Kotlin is designed with Java interoperability in mind. Existing Java code can be called from Kotlin in a natural way, and Kotlin code can be used from Java rather smoothly as well. In this section, we describe some details about calling Java code from Kotlin.

TheJeterLP:
> > That's why I didn't code stuff in kotlin yet. Because it's just bad to mix java with kotlin
> 
> From the official Kotlin docs:
> >Kotlin is designed with Java interoperability in mind. Existing Java code can be called from Kotlin in a natural way, and Kotlin code can be used from Java rather smoothly as well. In this section, we describe some details about calling Java code from Kotlin.
> 
> 

Yea, still not a fan about that. But it is what it is :)

TheJeterLP:
> no, this is the right way to do it in kotlin 

Okay, didn't know that. How does kotlin know what to expect from boolean? There's no java reference to that?

TheJeterLP:
> > How does kotlin know what to expect from boolean? There's no java reference to that?
> 
> I believe that kotlin derives the properties from the java class members. So it basically looks at the class, sees the `getBoolean()` function in a getter- pattern, and derives the property from it

okay, interesting. Whats if theres 2 boolean getters?

bogadana:
> Whats if theres 2 boolean getters?

then they have different names

shadow578:
> okay, interesting. Whats if theres 2 boolean getters?

kotlin will use the actual method name of the getter, not the type. In this case, it's confusing because the name __is__ literally boolean. 

say you have a java class like this:
```java
class MyClass {
  public boolean getFoo() {
     // ...
  }

  public void setFoo(boolean newFoo) {
     // ...
  }

  public double getBar() {
    // ...
  }
}
```

in kotlin, you'll be able to call the getters (and setters) just as if they were properties
```kotlin
val my = MyClass()

// this calls getFoo() 
val myBoolean = my.foo

// this calls setFoo(true)
my.foo = true

// and this calls getBar()
val myDouble = my.bar
```


The same is actually true in the other direction. so if you have a kotlin class with properties like this:
```kotlin
class MyClass(
  var foo: Boolean = false
)
```

in java, you'll see the property as a getter / setter pair and call like normal
```
MyClass my = new MyClass();
boolean foo = my.getFoo()
my.setFoo(true)
```


Hope this helps

shadow578:
> So could you get the Settings to work properly?

yep, turns out the `SettingsEnum` class just kinda ignored the default values instead of _actually_ using them. 
I think i'll have a look at the Settings system next, perhaps i can improve it a bit

oSumAtrIX:
Looks good to me, LGTM @TheJeterLP

PR #58, opened 2022-07-06
fix: Reduce padding DPI for swipe zones

EXAM454:
enable skip button by reducing padding dpi.


TheJeterLP (changes requested):
Please add the Settings back for this by using the existing setting in the SettingsEnum that was used by the swipe-controls. Would simplify some things.

PR #60, opened 2022-07-06
fix: change overlay time of swipe-controls

EXAM454:
change overlay time to 250ms or 500ms for fast volume & brightness panel switch. Also 2sec is iritating.


TheJeterLP:
> > Are you able to repro this behaviors?
> > 
> > [revanced/revanced-patches#124](https://github.com/revanced/revanced-patches/issues/124)
> 
> swipe to horizontally changes brightness i checked. but pause , skip , double tap skip working.

Pausing only works when you click on the middle of the screen tho. Not on the left or right space

PR #61, opened 2022-07-06
feat: `swipe-controls` rewrite

shadow578:
Basically full rewrite of fenster, this time with a (hopefully) better architecture that causes less issues. 
I'll add more details later


TheJeterLP:
You should merge upstream/main to your branch :)

Sculas:
Just curious, is Fenster the correct name? It just sounds like the original developer codes in their own language (German). So all references if Fenster should be changed to "Window", or a better name.

TheJeterLP:
> Just curious, is Fenster the correct name? It just sounds like the original developer codes in their own language (German). So all references if Fenster should be changed to "Window", or a better name.

I agree with that. Replace all Fenster with swipe-controls

shadow578:
> Just curious, is Fenster the correct name? It just sounds like the original developer codes in their own language (German). So all references if Fenster should be changed to "Window", or a better name.

i always assumed 'Fenster' to be a name for the swipe controls, like 'ReVanced' is a name for this project. 

> You should merge upstream/main to your branch :)

I believe i just broke something trying to do so, so i'll have to open a new PR later

PR #62, opened 2022-07-07
Update FensterController.kt

EXAM454:
FIX BUG: sometime volume brightness does not increse or decrese on swipe.


TheJeterLP:
Not going to merge this as swipe-controls feature is being rewritten in #64

PR #63, opened 2022-07-07
ENHANCEMENT:add transparancy effect to  black background of volume , brightness text.

EXAM454:
add transparancy effect to  black background of volume , brightness text.


TheJeterLP:
Not going to be merged as swipe-controls are currently being rewritten by the developer.

PR #81, opened 2022-07-13
feat: return youtube dislikes

j4k0xb:
related discussion: https://github.com/orgs/revanced/discussions/237, patches PR: https://github.com/revanced/revanced-patches/pull/175

It's now compatible with the new litho buttons (tested with frida hooks).
Updating the dislike count text after voting needs further research, but that can be added in a separate PR.


oSumAtrIX:
If we want to merge this, please create an issue for the dislike count text not updating after it has been merged.

oSumAtrIX (commented):
Why has so much code been deprecated?

TheJeterLP:
> > Why has so much code been deprecated?
> 
> Newer YT versions (tested on 17.14.35) use Litho instead of android Views for the dislike button so everything related to it can be removed.

Instead of deprecating that code should be removed then

TheJeterLP:
Please keep in mind to use a RYD class in the patches package which the actual patch can access then like all the other patches do :)

TheJeterLP:
Also please keep in mind to add the needed Settings to this from SettingsEnum

TheJeterLP:
Please fix the merge conflicts by merging upstream to your fork

PR #87, opened 2022-07-18
fix: make all patches toggleable with settings

TheJeterLP:
See tracking issue for more infos:
- https://github.com/revanced/revanced-patches/issues/197


OxrxL:
> > Some missing toggles here 👍
> > https://github.com/revanced/revanced-patches/blob/main/src/main/kotlin/app/revanced/patches/youtube/ad/general/bytecode/patch/GeneralBytecodeAdsPatch.kt
> 
> I'm not done yet :)

I know.

Just wanted to post it here as a memorandum (to help).

PR #88, opened 2022-07-18
fix: `swipe-controls` incompatible with `disable-fullscreen-panels`

shadow578:
closes https://github.com/revanced/revanced-patches/issues/196


TheJeterLP:
Could you please add an empty commit so we can squash? :)

shadow578:
Why is this needed? 

& does the commit message matter?

oSumAtrIX:
Empty commits are not required, you can squash a single commit.

PR #95, opened 2022-07-22
feat: `default-video-quality` patch

Canny1913:


app/src/main/java/app/revanced/integrations/patches/VideoQualityPatch.java
```diff
@@ -84,7 +82,8 @@ public static int setVideoQuality(Object[] qualities, int quality, Object qInter
         LogHelper.debug(VideoQualityPatch.class, "Index of quality " + quality + " is " + qualityIndex);
         try {
             Class<?> cl = qInterface.getClass();
-            Method m = cl.getMethod("x", Integer.TYPE);
+            Method m = cl.getMethod(qIndexMethod, Integer.TYPE);
```
TheJeterLP, line 85: Please try to not use reflection. Instead of doing it this way, return iStreamQualities.get(qualityIndex) and invoke the method in the patch. The integrations are only getters/setters. They are not intended to directly access youtube methods or variables.

TheJeterLP (changes requested):

app/src/main/java/app/revanced/integrations/patches/VideoQualityPatch.java
oSumAtrIX: IMO this is _fine_. Any other way is complicating it.

oSumAtrIX (approved):

app/src/main/java/app/revanced/integrations/patches/VideoQualityPatch.java
TheJeterLP: No its not fine. Integrations should never access YT methods directly. Thats not what the intention of the integrations is and no patch else does.

TheJeterLP: > IMO this is _fine_. Any other way is complicating it.

Its not over complicating. Its just returning a value and using it in a patch

TheJeterLP: Also this is very easy going to break as methodf "x" is probably going to be renamed. So an absolute no-go for me.

oSumAtrIX: ## 1.
> Integrations should never access YT methods directly. Thats not what the intention of the integrations is and no patch else does.

It is neither unintended nor does only this patch do it. There are a couple of patches that do it. You can either call the method through reflection OR modify the integrations through a patch to call the method OR inject the entire logic into the app. The latter is completely getting rid of the benefit of using the integrations and hence worse than doing it how @Canny1913 did.

## 2.
You suggested: 

> Instead of doing it this way, return iStreamQualities.get(qualityIndex) and invoke the method in the patch.

I said:

> OR inject the entire logic into the app. The latter is completely getting rid of the benefit of using the integrations which allow to simply code the entire logic and hence worse than doing it how @Canny1913 did.

And thus, your suggestion is overcomplicating and not true:

> Its not over complicating.

## 3.

> Also this is very easy going to break as method "x" is probably going to be renamed. So an absolute no-go for me.

The patch does nowhere use "x" anywhere. The patch is dynamically getting the method and thus is not going to break.

TheJeterLP: I've never seen any other patch calling a youtube method inside integrations. If thats the case, that should be changed. 

Its not overcomplicating, its keeping consistency around the code base. Patches edit and use youtube methods, integrations just provide helper functions such as settings. 

For "x" that was my mistake. I read that wrong since that was changed. I'm still not a fan of using reflection at all since its slow af too.

Canny1913: https://github.com/revanced/revanced-integrations/blob/main/app/src/main/java/app/revanced/integrations/patches/VideoSpeedPatch.java#L74

VideoSpeed invokes a method.

TheJeterLP: Integrations should be used the way a Library/Framework is used. And a library never calls a method from the project where its used. It just provides helper methods.

TheJeterLP: > https://github.com/revanced/revanced-integrations/blob/main/app/src/main/java/app/revanced/integrations/patches/VideoSpeedPatch.java#L74
> 
> VideoSpeed invokes a method.

I didn't see this. Should be changed too in my opinion.

TheJeterLP: The logic of the Patch should be in the Patches. Not inside of Integrations. Iterating though all youtube methods using reflection in integrations is not a good way to do it at all. Just saw that for VideoSpeed and I'm not a fan of that at all.

oSumAtrIX: > I've never seen any other patch calling a youtube method inside integrations

There is a couple patches as @Canny1913 noted.

> If thats the case, that should be changed.

Why.

> Its not overcomplicating.

I showed above why it is.

> Integrations just provide helper functions such as settings and LogHelper for example.

No, they do a lot more, they inject into views of YouTube, add listeners, modify parameters, and invoke functions. They are helper methods and should be used like that. I explained above why not invoking the method is worse.

> Integrations should be used the way a Library/Framework is used

No, you are now limiting the capabilities. This is an unnecessary restriction which requires to move logic to the patch.  
I said:

> they do a lot more, they inject into views of YouTube, add listeners, modify parameters, and invoke functions

By your ideals, they should do NONE of that.

TheJeterLP: Again, Reflection is SLOW. Very SLOW. It should not be used at compile time. Thats why doing something like this:
`for(Method m : class.getMethods()` is a no-go for java development. Reflection is a feature java has, but should be absolutely avoided to be used in all possible cases.

TheJeterLP: > Again, Reflection is SLOW. Very SLOW. It should not be used at runtime. Thats why doing something like this: `for(Method m : class.getMethods()` is a no-go for java development. Reflection is a feature java has, but should be absolutely avoided to be used in all possible cases.

oSumAtrIX: > Again, Reflection is SLOW

No, it is **relatively** slow and not slow.  The logger uses reflection the entire time.

>  It should not be used at compile time.

It is not being used at compile time?

> `for(Method m : class.getMethods()` is a no-go for java development

Please care to explain why? Additionally, this is nowhere used in the code. 

> Reflection is a feature java has, but should be absolutely avoided to be used in all possible cases.

So when should it be used according to you?

TheJeterLP: Reflection should never be used when its not absolutely needed. For ex. accessing private fields or methods. 
I changed my comment to runtime.
the for loop is used in the VideoSpeed patch, not in this one. But shit like that makes the youtube app slow. And that should be avoided at all cost because it will result in a bad user experience. More reflection == slower app == bad user xp.

Not using Reflection at all, and instead getting all the neccessary methods, fields, etc in the patch will make sure that the youtube apps response is and stays as fast as it is stock. Thats why it should be an absolute principle to not use reflection at all. Thats also an unspoken rule of general java development. ("If you can avoid it, do not use reflection.")

oSumAtrIX: In that case, if the patch becomes functional we should merge this and improve it afterward.

TheJeterLP: > In that case, if the patch becomes functional we should merge this and improve it afterward.

I could live with that. Buts it should be an absolute must to remove the use of reflection. Same goes for VideoSpeed patch and every other possible usage of it.

oSumAtrIX: Please open an issue regarding this.

TheJeterLP: I'm doing rn

app/src/main/java/app/revanced/integrations/patches/VideoQualityPatch.java
```diff
@@ -21,10 +21,16 @@ public class VideoQualityPatch {
     public static void changeDefaultQuality(int defaultQuality) {
         Context context = ReVancedUtils.getContext();
         if (isConnectedWifi(context)) {
-            SettingsEnum.PREFERRED_RESOLUTION_WIFI.setValue(defaultQuality);
+            SharedPreferences wifi = context.getSharedPreferences("revanced_prefs", 0);
+            SharedPreferences.Editor wifieditor = wifi.edit();
+            wifieditor.putInt("wifi_quality",defaultQuality);
+            wifieditor.apply();
```
oSumAtrIX, line 27: Please open an issue for this, as this is temporal

oSumAtrIX (changes requested):

app/src/main/java/app/revanced/integrations/patches/VideoQualityPatch.java
Canny1913: Opened one.

oSumAtrIX:
The files are unformated. Please format them next time before merging.

Canny1913:
> The files are unformated. Please format them next time before merging.

I will be creating a new PR anyway. Will do it next time 👍

PR #98, opened 2022-07-24
fix: simplify `old-quality` integration

OxrxL:
There's no other way to achieve a glitch-free `performItemClick`, neither with an onEndAnimation Listener (listView doesn't have any animation) for example, or other triggers. And we don't need a too much complex code.

Plus, despite not making the patch work in the absence of a delay (1ms is enough), runnable prevents FragmentManager crash issue in app.

This characteristic makes it also the safest way to apply this mod.

### DIFF
--- a/app/src/main/java/app/revanced/integrations/patches/OldQualityLayoutPatch.java
+++ b/app/src/main/java/app/revanced/integrations/patches/OldQualityLayoutPatch.java
@@ -1,52 +1,21 @@
 package app.revanced.integrations.patches;
 
-import android.graphics.Color;
-import android.graphics.drawable.ColorDrawable;
-import android.util.Log;
-import android.view.View;
-import android.view.ViewGroup;
-import android.view.Window;
+import android.os.Handler;
+import android.os.Looper;
 import android.widget.ListView;
-import android.widget.RelativeLayout;
-
-import java.util.Arrays;
 
 import app.revanced.integrations.settings.SettingsEnum;
 import app.revanced.integrations.utils.LogHelper;
 
 public class OldQualityLayoutPatch {
-    public static Window window;
-    private static boolean hideWindow = true;
-    
     public static void showOldQualityMenu(ListView listView)
     {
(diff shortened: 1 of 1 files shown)


app/src/main/java/app/revanced/integrations/patches/OldQualityLayoutPatch.java
```diff
@@ -1,52 +1,21 @@
...
+        listView.setVisibility(View.GONE);
 
-                parent.setVisibility(View.GONE);
-                if (window != null && hideWindow) {
-                    window.setBackgroundDrawable(new ColorDrawable(Color.TRANSPARENT));
-
-                    hideWindow = false;
+        new Handler(Looper.getMainLooper()).postDelayed(new Runnable() {
+                @Override public void run() {
+                        listView.performItemClick(null, 4, 0);
```
oSumAtrIX, line 17: This is not an elegant solution. It is error-prone as sleeping 1 millisecond might be too fast for example. Instead, when the child view is added, this event should be triggered.

oSumAtrIX (changes requested):

app/src/main/java/app/revanced/integrations/patches/OldQualityLayoutPatch.java
TheJeterLP: I agree on that. I'm going to close the PR, @OxrxL please reopen when you added those changes

OxrxL: I already explained why this solution is safe and glitch-free. respect the current one.

There's no other way to do that and obtain the same result.

OxrxL: > I agree on that. I'm going to close the PR, @OxrxL please reopen when you added those changes

There's no other way to get a satisfying solution, without this approach.

However, you are free to keep something that works badly.

PR #111, opened 2022-07-31
fix: app crashes on first start

TheJeterLP:
Currently when you migrate ReVanced from non-settings variant to settings-variant the app crashes because sharedprefs are handled different now. This currently results in the need for clearing app data. This PR fixes this.


Sculas:
Thanks!

PR #113, opened 2022-08-01
feat: `settings` patch framework 

oSumAtrIX:
RE:  https://github.com/revanced/revanced-patches/issues/158

Depends on:
- https://github.com/revanced/revanced-patches/pull/266


Sculas:
@oSumAtrIX You don't need to force push everytime next time. Just squash and merge, this way we can know what you did :)

oSumAtrIX:
> @oSumAtrIX You don't need to force push everytime next time. Just squash and merge after instead, this way we can know what you did :)

That was not possible since I had to mostly rebase when I was force-pushing.

Issue #118, opened 2022-08-08
bug: RYD doesnt work on new layout


j4k0xb:
### Type

Cosmetic

### Bug description

YouTube does A/B testing of a new layout/UI which causes RYD to break for some users.
- [x] ~~The like button displays the dislike count instead.~~ (fixed)
- [ ] The dislike button shows neither the initial text nor the dislike count.

### Steps to reproduce

Cant, still shows the old layout for me

### Relevant log output
Debug logs provided by Bleuzen

old layout:
```shell
[...]

PR #119, opened 2022-08-08
fix: unintentional like button detection

j4k0xb:
Partially fixes
- #118 

will continue to work on the old layout but at least the likes on the new one won't get overwritten.


oSumAtrIX:
Thanks

PR #122, opened 2022-08-18
feat: `overlay-buttons` patch

inotia00:
I didn't know what to name this, so I named it 'overlay-buttons'
This is a patch for a button inside the player that was in Vanced. (Copy, Copy With Timestamp, Autorepeat)

Screenshot
https://imgur.com/a/m3HLV2G

Since the button container at the bottom of the player has been moved to litho, instead of using the existing array settings (NONE, PLAYER, BUTTON_CONTAINER, BOTH), I simply implemented it as an ON/OFF switch (BOOLEAN).

needs to be merged together with [revanced-patches](https://github.com/revanced/revanced-patches/pull/353)


inotia00:
Later, when I merge the branch according to the latest version, I will leave a new PR at that time.
For now, I will close this PR.

PR #128, opened 2022-08-26
feat: long press gesture toggle for swipe or native gesture

LoV432:
https://github.com/orgs/revanced/discussions/541


oSumAtrIX (approved):

oSumAtrIX:
@LoV432 after this PR, is it still possible to use swipe gestures without long pressing at all?

LoV432:
> @LoV432 after this PR, is it still possible to use swipe gestures without long pressing at all?

I dont see the reason for pr anymore after https://github.com/revanced/revanced-integrations/pull/123 is merged.

PR #129, opened 2022-08-27
fix: pull up to refresh

OxrxL:
Still need to be tested further, to be sure it always works.


oSumAtrIX:
Nothing is being added to the new array, what is the plan?

OxrxL:
> Nothing is being added to the new array, what is the plan?

The plan is to replace the string called "value" with the bytebuffer. In this way we should be able to block literally everything related to ads elements (also the empty containers that "value" variable can't detect).

oSumAtrIX:
Îs `value` still required? If not, please remove it.

OxrxL:
Obviously we need to keep "value" variable for searching further new ads. It's better of bytebuffer for debug.

dibyarupprk:
It doesn't work, still pull down to refresh is buggy

PR #134, opened 2022-09-07
feat: disable sponsorblock on shorts

OxrxL:


oSumAtrIX:
Please create separate PRs for both changes so they can be squashed separately.

PR #139, opened 2022-09-15
fix: add more block string to `LithoAd` filter

d4rkk3y:
Add `video_display_full_layout` to value blockList of LithoAd filter. Possibly fix https://github.com/revanced/revanced-patches/issues/432


oSumAtrIX:
Thanks, further testing will conclude if it works.

PR #140, opened 2022-09-15
fix: more generalized `litho ads`removal

OxrxL:
This is a different approach to ads blocking, and it's based on a simple concept:

"Don't block the template by its name, but by its link"

Every ad have at least one link, from which get the informations shown to the user, and this link have a unique pattern (3 in total, for all ad).


OxrxL:
@oSumAtrIX Try this too.

oSumAtrIX:
The ones you removed, did you account for those?

OxrxL:
> The ones you removed, did you account for those?

Are all condensed in those 3 values.

PR #142, opened 2022-09-16
feat(tiktok): `tiktok-feed-filter` , `tiktok-settings` and `tiktok-force-login`

d4rkk3y:


oSumAtrIX:
Soon, every patch will have their own integrations. General integrations like this will also still exist providing basic utility methods like settings or methods to the Android API (which the patch integrations will not have access to), but patch specific code will be moved to the patches own integrations. The current problem with this PR is that using this integrations in other applications will still merge the integrations code for TikTok patches. Would it be possible to make this repository build separate integrations, in this case, one for TikTok and the other for YouTube? This way, we can then easily migrate to patch specific integrations and prevent unrelated code from getting merged.

Sculas:
@oSumAtrIX I think that the way this PR is, is okay for now. It's better to just have it be "wrong" and then fix it later than doing a half-assed implementation and having to fix it later anyway. But it's up to you.

oSumAtrIX:
Actually, merging unrelated code is more of a half assed implementation then creating a proper solution which would be integrations specifically made for each patch, or until thats done, for these patches for now. This way, useless code would be prevented to be merged to our current apps and the same for the app @d4rkk3y patches. The solution is simply to create a separate integrations apk. Then either you use the current integrations or the new ones specifically for these patches.

Sculas:
That'd require a whole new repository and new IntegrationsPatch dependency for.. just this? And for it then all be removed when everything is refactored into how you said it would be like? Don't think only about yourself if that's what you do, think about the author too who has to do all of that with no added value.

d4rkk3y:
@oSumAtrIX @Sculas I have a idea. How about create a annotation to mark a class in integration belong to target app? And then implement merge method in `patcher` which will check it and merge if current patching app is target app.

oSumAtrIX:
> That'd require a whole new repository and new IntegrationsPatch dependency for.. just this? And for it then all be removed when everything is refactored into how you said it would be like? Don't think only about yourself if that's what you do, think about the author too who has to do all of that with.
> 
> Instead, if you insist on going with this route, merge this PR and make that huge change yourself or ask if the author would like to do that. Either way, you'll have to help because you need to make a repository for it.

No, that's not required, it can be done in this repo pretty straight forward. Less refactoring will happen if the integrations are being started to be separated now for each app.

E85Addict:
I've tested both this and [#501](https://github.com/revanced/revanced-patches/pull/501) together and it seems to work flawless, is there anything left that is required in the PR's before merge?

oSumAtrIX:
> @oSumAtrIX @Sculas I have a idea. How about create a annotation to mark a class in integration belong to target app? And then implement merge method in `patcher` which will check it and merge if current patching app is target app.

This would require to annotate all classes which should be merged, including those which are shaded into the integrations to which an annotation can not be added. As an alternative every class which does not have this particular annotation, could be merged by default, while those who have, but do not match the selected app, are excluded. For the scope of this PR it is better to separate this to another PR, please revert the intruduction of the `MergeIf` annotation so we can merge this PR for now.

app/src/main/java/app/revanced/tiktok/settings/SettingsEnum.java
```diff
@@ -0,0 +1,149 @@
+package app.revanced.tiktok.settings;
+
+import android.content.Context;
+import android.util.Log;
+
+import app.revanced.tiktok.utils.LogHelper;
+import app.revanced.tiktok.utils.ReVancedUtils;
+import app.revanced.tiktok.utils.SharedPrefHelper;
+import merger.MergeIf;
```
oSumAtrIX, line 9:
```suggestion
```

app/src/main/java/app/revanced/tiktok/settings/ReturnType.java
```diff
@@ -0,0 +1,7 @@
+package app.revanced.tiktok.settings;
+
+import merger.MergeIf;
```
oSumAtrIX, line 3:
```suggestion
```

app/src/main/java/app/revanced/tiktok/feedfilter/FeedItemsFilter.java
```diff
@@ -0,0 +1,40 @@
+package app.revanced.tiktok.feedfilter;
+
+import com.ss.android.ugc.aweme.feed.model.Aweme;
+import com.ss.android.ugc.aweme.feed.model.FeedItemList;
+
+import java.util.Iterator;
+import java.util.List;
+
+import app.revanced.tiktok.settings.SettingsEnum;
+import merger.MergeIf;
```
oSumAtrIX, line 10:
```suggestion
```

app/src/main/java/app/revanced/tiktok/MergeIfPackage.java
```diff
@@ -0,0 +1,7 @@
+package app.revanced.tiktok;
+
+import merger.MergeIf;
+
+@MergeIf(packageName = {"com.ss.android.ugc.trill", "com.zhiliaoapp.musically"})
+public class MergeIfPackage {
+}
```
oSumAtrIX, line 7: Remove this file.

app/src/main/java/app/revanced/integrations/MergeIfPackage.java
```diff
@@ -0,0 +1,7 @@
+package app.revanced.integrations;
+
+import merger.MergeIf;
+
+@MergeIf(packageName = {"com.google.android.youtube"})
+public class MergeIfPackage {
+}
```
oSumAtrIX, line 7: Remove this file.

oSumAtrIX (changes requested):

E85Addict:
Tested the corrections all is working!

oSumAtrIX (approved):

PR #143, opened 2022-09-17
fix(sponsorblock): return details of 400 error

mchangrh:
Users are coming into the SponsorBlock discord presenting error 400, which is a generic catch-all for inputs that are bad or impossible. This is usually a bad userID but eating the error makes troubleshooting extremely difficult when it's not.

Possible causes of 400
- userID
- videoID
- no segments
- invalid segment times
- descriptions on non-chapters
- chapters that exceed limit
- lack of permissions


oSumAtrIX:
The new string with that id has to be added to the strings resource in the patches too. Looks good other than that.

mchangrh:
> The new string with that id has to be added to the strings resource in the patches too. Looks good other than that.

https://github.com/revanced/revanced-patches/pull/512

Does this cover the string resources or is there somewhere else as well

PR #147, opened 2022-09-21
feat(tiktok): signature spoof

d4rkk3y:


oSumAtrIX:
To make this patch smaller, would it not be possible to patch the necessary call to the Android API? Right now you reflectively modify the call to it.

d4rkk3y:
@oSumAtrIX did you mean only patch the call that cause the issue?

oSumAtrIX:
No, this api is used in TikTok. Finding where and patching that would allow to get rid of the entire reflective workaround.

d4rkk3y:
> No, this api is used in TikTok. Finding where and patching that would allow to get rid of the entire reflective workaround.

Ah I see. I will try that way. I hope TikTok did not call this api in native lib.

oSumAtrIX:
Probably not, if they were reflection would've surely failed too.

d4rkk3y:
Yes, If they check apk directly reflection will fail too. But if they call Android API from native lib it will not.

oSumAtrIX:
What i meant, the reflection proxy succeeded, so it's likely not in a native library.

d4rkk3y:
It seem TikTok don't not check signature use android api. I tested again and this did not work.

PR #153, opened 2022-09-23
fix: remove distributionSha256Sum tag

OxrxL:
It cause sync error on Android Studio


oSumAtrIX:
Explain sync error.

OxrxL:
> Explain sync error.

Immagine 2022-09-23 110454

PR #158, opened 2022-09-26
feat: enable minification 

Canny1913:
Increases merging speeds by decreasing the amount of kotlin classes.


PalmDevs:
CANNY IS POG

PR #168, opened 2022-10-04
fix(Remember Video Quality): remember higher qualities

johnconner122:
closes:
https://github.com/revanced/revanced-patches/issues/719


oSumAtrIX:
Can you explain, how it solves the related issue?

PR #187, opened 2022-10-26
fix(YouTube/general ads): alphabetical reorder

OxrxL:


OxrxL:
> Some orders are chosen intentionally. We expect to see occurences of some strings earlier, so putting them early in the list helps with performance, sorting variables is fine, but do not change the order of existing lists.

Can i close the PR so?

PR #189, opened 2022-10-26
feat: hide full screen live chat button

OxrxL:


app/src/main/java/app/revanced/integrations/patches/HideLiveChatFullScreenButtonPatch.java
```diff
@@ -0,0 +1,17 @@
...
+
+import app.revanced.integrations.settings.SettingsEnum;
+
+public class HideLiveChatFullScreenButtonPatch {
+    //Used by app.revanced.patches.youtube.layout.comments.patch.CommentsPatch
+    public static int fullScreenLiveChatButtonId = 0;
+
+    public static int hideLiveChatFullScreenButton(View view, int originalValue) {
+        if (!SettingsEnum.HIDE_LIVE_CHAT_FULL_SCREEN_BUTTON.getBoolean()) return originalValue;
+        else if (view.getId() == fullScreenLiveChatButtonId) return 8;
```
oSumAtrIX, line 13: Why is this required?

oSumAtrIX (changes requested):

app/src/main/java/app/revanced/integrations/patches/HideLiveChatFullScreenButtonPatch.java
OxrxL: Because to check which view to hide, I need the wide literal resource id of the live chat button.

The class where this patch is placed, is a class shared between many views (all the buttons of fullscreen player). So...I can easily hide the right view with an equal check condition between the live chat button resource id, and the resource id of view currently managed by the class (as you can see from integration code).

app/src/main/java/app/revanced/integrations/patches/HideLiveChatFullScreenButtonPatch.java
```diff
@@ -0,0 +1,17 @@
...
+
+        return originalValue;
```
oSumAtrIX, line 15:
```suggestion
        if (!SettingsEnum.HIDE_LIVE_CHAT_FULL_SCREEN_BUTTON.getBoolean() || view.getId() != fullScreenLiveChatButtonId) return originalValue;

        return 8;
```

oSumAtrIX (changes requested):

oSumAtrIX (approved):

PR #190, opened 2022-10-28
Change that auto skipped will only trigger once per segment

thebestnom:
Should work, now testing

### DIFF
--- a/app/src/main/java/app/revanced/integrations/sponsorblock/PlayerController.java
+++ b/app/src/main/java/app/revanced/integrations/sponsorblock/PlayerController.java
@@ -183,7 +183,7 @@ public class PlayerController {
                 continue;
 
             // we are in the segment!
-            if (segment.category.behaviour.skip) {
+            if (segment.category.behaviour.skip && !(segment.category.behaviour.key.equals("skip-once") && segment.didAutoSkipped)) {
                 sendViewRequestAsync(millis, segment);
                 skipSegment(segment, false);
                 break;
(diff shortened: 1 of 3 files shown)


oSumAtrIX:
Is this WIP as the title suggests, if so please mark this PR as a draft.

thebestnom:
Sorry, for some reason thought that writing [WIP] will auto draft it 😅 (it should work like that, but I do want to test first)

thebestnom:
Ok, it half works, it does allow to watch second time, but the skip button doesn't work...will complete this tomorrow

thebestnom:
Ok, now first skip works and skip button does work on second time, but overlapping segment it skips only the first segment

thebestnom:
Ok, now it works

oSumAtrIX:
Can you explain all changes and the new behaviour?

thebestnom:
The new behavior is that if a segment did auto skipped, it will do it just once, so if you go back to the same segment now the skip button will be there instead of skipping

The changes are that now segment object have a skipped property to save if it already did auto skipped
Than the skip button will check if the skip and that it didn't auto skipped to see if it needed to be shown
And added that the skip function will return if it succeed or not so I could check if the segment did actually skipped or not so it could put the auto skipped  property correctly only if it did skip

thebestnom:
example
https://user-images.githubusercontent.com/21955247/198844214-4eb4652e-5826-4fea-9e21-848d37ec75fd.mp4

oSumAtrIX:
If possible, can you add a settings switch which enables back the defaut behaviour? Something like "Skip once". If the switch is turned off (by default it should be on), then it will skip the segment every time.

thebestnom:
Can try, but didn't touch patches ever so it would take me more time 😅 (but if Ill do I will do that you could enable it per segment type)

oSumAtrIX:
You would only need to make changes to the integrations.
RE: https://github.com/revanced/revanced-integrations/blob/d745e293953e61f01970ccdc6937572e25744146/app/src/main/java/app/revanced/integrations/sponsorblock/SponsorBlockSettings.java#L99

thebestnom:
Works but the string is skip_automattically_once, how do we add to strings?

thebestnom:
Added string to patcces, didn't test

thebestnom:
I did test the new enum and it works, didn't test the strings.xml patch

thebestnom:
https://github.com/revanced/revanced-patches/pull/907 Works

oSumAtrIX (approved):

oSumAtrIX:
Thanks

PR #193, opened 2022-10-30
fix(youtube/general-ads): remove text banner ads

inotia00:
In the past, I have left the same PR. https://github.com/revanced/revanced-integrations/pull/179
However, over time I got reports from testers that text banner ads were still displayed.

Unfortunately, text banner ads did not display on my device.
So I've been testing text banner blocking on people who run into issues using several litho filters.

As a result, I was able to infer a filter that blocks text banners.

After testing for about a week, no one saw the text banner ads reappear.


oSumAtrIX:
Can you attach an image for text banners?

inotia00:
photo_2022-10-28_16-39-25
this is reported screenshot

oSumAtrIX:
If I recall correctly the path builder of this type of ad is indistinguishable from normal video components until it is too late to remove the text component which remains visible.

PR #202, opened 2022-11-13
fix(youtube/theme): shorts comment box color

OxrxL:
Closes: https://github.com/revanced/revanced-patches/issues/922


OxrxL:
In practice: not all UI elements can parse color from "0" value. Some of them needs to get the same color from different value (-16777215 is equal to -0xFFFFFF, that is the negation of full white color).

app/src/main/java/app/revanced/integrations/patches/LithoThemePatch.java
```diff
@@ -24,7 +24,7 @@ public static int applyLithoTheme(int originalValue) {
         var isDarkTheme = ThemeHelper.isDarkTheme();
         
         if ((isDarkTheme && anyEquals(originalValue, DARKCONSTANTS)) || (!isDarkTheme  && anyEquals(originalValue, WHITECONSTANTS)))
-                return 0;
+                return -16777215;
```
oSumAtrIX, line 27: What does this number represent

oSumAtrIX:
Thanks

PR #203, opened 2022-11-13
feat(youtube/theme): explore drawer background

OxrxL:
# TODO:

- [x] Find the constant for light mode

https://www.reddit.com/r/revancedapp/comments/ypt7ii/black_theme_doesnt_get_applied_here_in_the/

### DIFF
--- a/app/src/main/java/app/revanced/integrations/patches/LithoThemePatch.java
+++ b/app/src/main/java/app/revanced/integrations/patches/LithoThemePatch.java
@@ -13,6 +13,7 @@ public class LithoThemePatch {
     };
 
     private static final int[] DARKCONSTANTS = {
+        -14145496, // 'Explore' drawer background
         -14606047, // comments chip background
         -15198184, // music related results panel background
         -15790321, // comments chip background (new layout)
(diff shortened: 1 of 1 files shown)


app/src/main/java/app/revanced/integrations/patches/LithoThemePatch.java
```diff
@@ -13,6 +13,7 @@ public class LithoThemePatch {
     };
 
     private static final int[] DARKCONSTANTS = {
+        -14145496, // 'Explore' drawer background
```
oSumAtrIX, line 16:
```suggestion
        -14145496, // explore drawer background
```

oSumAtrIX (commented):
What about light mode?

OxrxL:
> What about light mode?

This value was found by inotia. I couldn't, because I didn't have this panel (and I still don't have it now).

inotia00:
I also tested it in light mode
there is no litho theme filter to add for light mode

this is the screenshot when i changed the `lightThemeBackgroundColor` to pink

ReVanced2

PR #204, opened 2022-11-13
refactor(youtube/general-ads-patch): improved exclusion blocking

OxrxL:
Previous exclusion blocking didn't exclude all components of the comments section (such as buttons, textboxes etc.), slowing down the application when scrolling through many comments. This refactor allows to reduce not only the number of values needed for blocking, but (such on the previous code) also avoid the unwanted ```history button``` and ```download button``` blockages.


OxrxL:
@oSumAtrIX This should fix a bug related to a crash, that I got after several comments (and answers) scrolling.

Can you merge it asap?

app/src/main/java/app/revanced/integrations/patches/LithoFilterPatch.java
```diff
@@ -214,7 +214,7 @@ public GeneralBytecodeAdsPatch() {
                 //"full_width_square_image_layout",
                 "video_display_full_buttoned_layout",
                 "_ad",
-                "ad_",
+                "|ad_",
```
oSumAtrIX, line 217: What about components called something_ad_something?

app/src/main/java/app/revanced/integrations/patches/LithoFilterPatch.java
```diff
@@ -258,16 +258,11 @@ public GeneralBytecodeAdsPatch() {
     public boolean filter(final String path, final String identifier) {
         // Do not block on these
         if (ReVancedUtils.containsAny(path,
+                "comment_thread",
```
oSumAtrIX, line 261: What is this?

app/src/main/java/app/revanced/integrations/patches/LithoFilterPatch.java
```diff
@@ -258,16 +258,11 @@ public GeneralBytecodeAdsPatch() {
...
                 "related_video_with_context",
                 "search_video_with_context",
-                "download_",
-                "library_recent_shelf",
-                "menu",
-                "root",
-                "-count",
-                "-space",
-                "-button"
+                "horizontal_shelf"
```
oSumAtrIX, line 265: I remember couple ad components which are "horizontal_shelf". Do you consider that?

oSumAtrIX (changes requested):

app/src/main/java/app/revanced/integrations/patches/LithoFilterPatch.java
OxrxL: The name used by comments component (when you open comment section and scroll comments).

app/src/main/java/app/revanced/integrations/patches/LithoFilterPatch.java
OxrxL: This is the name used by history button components.

So, excluded it lead to the entire history button exclusion.

app/src/main/java/app/revanced/integrations/patches/LithoFilterPatch.java
OxrxL: You explicitly excluded download_button. But it's useless, because you can add a | on ```ad_```, to to identify ads components only instead of other components.

app/src/main/java/app/revanced/integrations/patches/LithoFilterPatch.java
OxrxL: > I remember couple ad components which are "horizontal_shelf". Do you consider that?

Anyway...you remember horizontal_video_shelf, that is another thing.

app/src/main/java/app/revanced/integrations/patches/LithoFilterPatch.java
oSumAtrIX: That does not answer the question. The block rule `generalAds` will not consider components called something_ad.

OxrxL: There are two ads whose name starts with ad_

So _ad was added as a valid name for both, avoiding adding two different names.

oSumAtrIX: And what components with the name "something_ad"?

OxrxL: > And what components with the name "something_ad"?

Mhh...seems there is just one:

Screenshot_20221113_190854_Samsung Internet

OxrxL: So we need:

_ad
_ad_
ads_video_with_context

app/src/main/java/app/revanced/integrations/patches/LithoFilterPatch.java
```diff
@@ -214,7 +214,7 @@ public GeneralBytecodeAdsPatch() {
                 //"full_width_square_image_layout",
                 "video_display_full_buttoned_layout",
                 "_ad",
-                "ad_",
+                "ad_badge",
```
oSumAtrIX, line 217: What about components named ad_something?

OxrxL: There are no ad_something based on repo history (only ad_badge).

There was just ```text_search_ad``` and ```text_display_ad```. But for these can be used ```_ad``` instead ```ad_```

OxrxL: I posted a screenshot with all ads template names, in the previous conversation.

oSumAtrIX (approved):

PR #205, opened 2022-11-13
fix(youtube/general-ads-patch): improved blocking furthermore

OxrxL:
To avoid eventually other unwanted blocks, it would be better to perform a more selective ads blocking, also because the isn't any template name that ends with ```_ad``` (instead there are 2 of that which contains ```_ad_```).


oSumAtrIX:
There were components called something_ad.

OxrxL:
> There were components called something_ad.

Sure?

PR #215, opened 2022-11-19
fix(youtube/general-ads): block ad links

aliernfrog:
This fixes random ad links appearing in feed.
IMG_20221119_153655


oSumAtrIX:
Is this tested? Another PR was claiming a similar fix, but it was discussed, that this components pathBuilder did not have this string. If it fixes it, please remove the other filter, which just partially blocks the ad component. If both are needed, please add a seperate `BlockRule` with the same condition to be enabled as the general-ads `BlockRule`.

oSumAtrIX:
Closing in favour of 2c8b23426f88111646cd0513517feff7a65a7c13.

PR #222, opened 2022-11-24
logging improvements

LisousEinaiKyrios:
logging changes

PR #224, opened 2022-11-29
ReturnYouTubeDislike - handle if YouTube video owner set video likes to hidden

LisousEinaiKyrios:
RYD API does not collect or provide data for videos where the creator has enabled hiding the like count.

Example video: https://www.youtube.com/watch?v=UnrU5vxCHxw
RYD data: https://returnyoutubedislikeapi.com/votes?videoId=UnrU5vxCHxw

As per: https://github.com/Anarios/return-youtube-dislike/discussions/530 
Do not show a dislike count for these type of videos, as no dislike data is available.

Other code changes include:
- use HTTP request headers as specified by RYD API
- added a thread pool executor (video UI seems to load a little faster since UI does not wait for a resource heavy thread to create and start)
- simplified ReturnYouTubdeDislike fetch code using a Future object
- removing the non-functional interrupt logic (If the calling thread is interrupted, HttpURLConnection continues to run and does not halt or throw an exception).  Interrupting is not needed anyways since the timeout is so short.
- voting runs on a single thread executor (removes the need to synchronize voting/registration)
- reduced UI blocking timeout to 4 seconds (with 5 seconds it could still trigger an Android "App Not Responding" message)

I've been using this for a few days and it's working well for me.


oSumAtrIX:
#222 caused some merge conflicts and updated how logging works, before merging, this should be looked into.

LisousEinaiKyrios:
It's good to go

app/src/main/java/app/revanced/integrations/requests/Requester.java
```diff
@@ -24,32 +25,52 @@ public static HttpURLConnection getConnectionFromRoute(String apiUrl, Route rout
         return connection;
     }
 
+    /**
+     * Parses the connection stream, and then and disconnects the {@link HttpURLConnection}
+     */
```
oSumAtrIX, line 30: Comments are a inconsistent in style, please make sure to use one style, preferably, the same as existing comments.

app/src/main/java/app/revanced/integrations/returnyoutubedislike/ReturnYouTubeDislike.java
```diff
@@ -45,172 +67,118 @@ public enum Vote {
...
             var conversionContextString = conversionContext.toString();
 
+            boolean isSegmentedButton = false;
             // Check for new component
             if (conversionContextString.contains("|segmented_like_dislike_button.eml|")) {
-                segmentedButton = true;
+                isSegmentedButton = true;
             } else if (!conversionContextString.contains("|dislike_button.eml|")) {
-                LogHelper.printDebug(() -> "could not find a dislike button in " + conversionContextString);
+                LogHelper.printDebug(() -> "ignoring UI component: " + conversionContextString);
```
oSumAtrIX, line 126:
```suggestion
```

app/src/main/java/app/revanced/integrations/returnyoutubedislike/ReturnYouTubeDislike.java
```diff
@@ -222,48 +190,82 @@ public static void sendVote(Vote vote) {
...
+                // example video: https://www.youtube.com/watch?v=UnrU5vxCHxw
+                // RYD data: https://returnyoutubedislikeapi.com/votes?videoId=UnrU5vxCHxw
+                //
+                // discussion about this: https://github.com/Anarios/return-youtube-dislike/discussions/530
+
+                //
+                // RYD browser plugin shows the dislikes as a localized translated string of "disabled by owner"
+                // ReVanced could do the same, but for now just leave as-is
+                //
+                return false;
```
oSumAtrIX, line 236: If returned here, the user would not get any feedback on wether the dislikes are disabled, the server is down, or wether the patch has an issue. Some kind of feedback should be given such as the string in the comment above, or something shorter such as "OFF"

oSumAtrIX (changes requested):

app/src/main/java/app/revanced/integrations/returnyoutubedislike/ReturnYouTubeDislike.java
LisousEinaiKyrios: I changed the "Likes" string to say:
   "Likes Hidden by Creator"

It makes the button somewhat large, but it more accurately describes the situation.
These type of videos are not common, so I don't think the extra UI space is not a big concern.

app/src/main/java/app/revanced/integrations/requests/Requester.java
LisousEinaiKyrios: I cleaned up the comments a bit

app/src/main/java/app/revanced/integrations/returnyoutubedislike/ReturnYouTubeDislike.java
oSumAtrIX: Shouldn't it say "dislikes hidden by creator"? (I do not remember, but if the creator hides the dislike count, is that already written in the likes string?)

LisousEinaiKyrios: The only indication that likes are hidden by the creator, is no number of likes is shown.
If the 'Likes' label says "Dislikes Hidden by Creator", but then users might question "But where is the likes count?"

If the [cosmetic layout bug](https://github.com/revanced/revanced-integrations/issues/118) is resolved, then I think the ideal is to show "Dislikes Hidden by Creator" in the 'Dislike' label.  But right now that's not possible since both likes and dislikes are jammed into the single 'Likes' label.

Maybe show "Likes and Dislikes Hidden by Creator"?
It's very wordy, but it would completely convey the situation.

LisousEinaiKyrios: "Likes and Dislikes Hidden by Creator" is too wordy, and the dislike button is almost off the screen.

LisousEinaiKyrios: [Likes Hidden by Creator](https://freeimage.host/i/HfAHiVR)

[Likes and Dislikes Hidden by Creator](https://freeimage.host/i/HfAHsPp)

oSumAtrIX: Can the creator even "only" hide the likes?

LisousEinaiKyrios: The creator can only hide the likes.  

Dislikes are always hidden by YouTube.

oSumAtrIX: In that case, just saying "Hidden" would suffice while also be short enough to not be obnoxiously wide.

app/src/main/java/app/revanced/integrations/returnyoutubedislike/ReturnYouTubeDislike.java
```diff
@@ -219,9 +219,9 @@ private static void updateDislike(AtomicReference<Object> textRef, boolean isSeg
                 // discussion about this: https://github.com/Anarios/return-youtube-dislike/discussions/530
 
                 //
-                // Change the "Likes" string to show that likes are hidden
+                // Change the "Likes" string to show that likes and dislikes are hidden
                 //
-                newDislikeString = "Likes Hidden by Creator"; // for now, this is not localized
+                newDislikeString = "Likes Hidden"; // for now, this is not localized
```
oSumAtrIX, line 224: Not only the hides would be hidden, also the dislikes, which is why this should say "Hidden" only.

oSumAtrIX (changes requested):

app/src/main/java/app/revanced/integrations/returnyoutubedislike/ReturnYouTubeDislike.java
LisousEinaiKyrios: changed to "Hidden"

oSumAtrIX (approved):

PR #238, opened 2022-12-18
refactor(youtube/general-ads-patch): added new names

0xrxL:
"search_video_with_context" are accessible through the opening of the following components:

Screenshot_20221218_224551

Meanwhile the component "|comment." exists only on comments replies, in place of "comment_thread".

### DIFF
--- a/app/src/main/java/app/revanced/integrations/patches/GeneralAdsPatch.java
+++ b/app/src/main/java/app/revanced/integrations/patches/GeneralAdsPatch.java
@@ -11,7 +11,9 @@ public final class GeneralAdsPatch extends Filter {
     private final String[] IGNORE = {
             "home_video_with_context",
             "related_video_with_context",
+            "search_video_with_context",
             "comment_thread", // skip blocking anything in the comments
+            "|comment.", // skip blocking anything in the comments replies
             "download_",
             "library_recent_shelf",
             "playlist_add_to_option_wrapper" // do not block on "add to playlist" flyout menu
--- a/app/src/main/java/app/revanced/integrations/patches/GeneralAdsPatch.java
+++ b/app/src/main/java/app/revanced/integrations/patches/GeneralAdsPatch.java
@@ -11,7 +11,6 @@ public final class GeneralAdsPatch extends Filter {
     private final String[] IGNORE = {
             "home_video_with_context",
             "related_video_with_context",
-            "search_video_with_context",
             "comment_thread", // skip blocking anything in the comments
             "|comment.", // skip blocking anything in the comments replies
             "download_",


oSumAtrIX:
These are already blocked. On which query do these appear?

0xrxL:
> These are already blocked. On which query do these appear?

As I said: ```search_video_with_context``` appears on the chapters list above video card

Screenshot_20221218_225245

meanwhile ```|comment.``` appears on comments replies ("risposte" in italian 👍).

Screenshot_20221218_225255

oSumAtrIX:
Please explain this PR in full detail so I can properly follow what this PR does. I can not understand the motivation behind this PR. Chapter teaser are hidden with 
https://github.com/revanced/revanced-integrations/blob/04a89af1c9d5d59ca88b842a2a4d1301703e3c99/app/src/main/java/app/revanced/integrations/patches/GeneralAdsPatch.java#L39

0xrxL:
> Please explain this PR in full detail so I can properly follow what this PR does. I can not understand the motivation behind this PR. Chapter teaser are hidden with
> 
> https://github.com/revanced/revanced-integrations/blob/04a89af1c9d5d59ca88b842a2a4d1301703e3c99/app/src/main/java/app/revanced/integrations/patches/GeneralAdsPatch.java#L39

This PR adds 2 new items to the Ignore list, in order to avoid useless contains checks on the string name of these...and also to avoid unwanted blocking of components (like the one of the playlist, if you remember).

0xrxL:
At the moment if you open the replies comments panel, the components present in it are not ignored and therefore performance slowdowns can occur, because the code checks if |comment. can be blocked.

But there's no point to block comments replies. So it's worth adding this component to the Ignore list.

oSumAtrIX:
Makes sense, what about
https://github.com/revanced/revanced-integrations/blob/1921538f3f64ef13ec836f10b621690f67cdf1f8/app/src/main/java/app/revanced/integrations/patches/GeneralAdsPatch.java#L15

What is the purpose of adding `|comment.` in addition to that? And what about `search_video_with_context`?

0xrxL:
> Makes sense, what about
> 
> https://github.com/revanced/revanced-integrations/blob/1921538f3f64ef13ec836f10b621690f67cdf1f8/app/src/main/java/app/revanced/integrations/patches/GeneralAdsPatch.java#L15
> 
> What is the purpose of adding `|comment.` in addition to that? And what about `search_video_with_context`?

```comment_thread``` is the name of root components of comments list (not replies).

```search_video_with_context``` is the root name of components of this UI panel (as I explained in the previous comment):

Screenshot_20221218_225245

**Comment by 0xrxL** [comment]
@oSumAtrIX Can be merged?

**Comment by oSumAtrIX** [comment]
If you add `search_video_with_context` it is not possible to block chapter previews.

**Comment by oSumAtrIX** [comment]
As explained [here](https://github.com/revanced/revanced-integrations/pull/238#issuecomment-1356884779)

**Comment by 0xrxL** [comment]
@oSumAtrIX ```search_video_with_context``` has been removed with last commit. 👍

File: app/src/main/java/app/revanced/integrations/patches/GeneralAdsPatch.java [hunk 1]
```diff
@@ -12,6 +12,7 @@ public final class GeneralAdsPatch extends Filter {
             "home_video_with_context",
             "related_video_with_context",
             "comment_thread", // skip blocking anything in the comments
+            "|comment.", // skip blocking anything in the comments replies
```
oSumAtrIX, line 15: What do the pathBuilder strings look like for ˋcomment_threadˋ and ˋ|commentˋ?

oSumAtrIX (approved):

app/src/main/java/app/revanced/integrations/patches/GeneralAdsPatch.java
0xrxL: I added ```|comment.``` to define a component name that contains only a reference like that:

```CellType|comment.eml```

In this way, despite the component name doesn't contain ```comment_thread```, it can be ignored as well

0xrxL: > What do the pathBuilder strings look like for ˋcomment_threadˋ and ˋ|commentˋ?

For comments reolies, something like that:

```CellType|comment.eml```

PR #239, opened 2022-12-18
fix(youtube/theme): shorts comment box wrong color

0xrxL:
Closes: https://github.com/revanced/revanced-patches/issues/922


app/src/main/java/app/revanced/integrations/patches/GeneralAdsPatch.java
```diff
@@ -102,6 +103,8 @@ public GeneralAdsPatch() {
     public boolean filter(final String path, final String identifier) {
         BlockResult result;
 
+        shortsCommentBox = path.contains("comment_composer");
```
oSumAtrIX, line 106: This has nothing to do in the `GeneralAdsPatch` class. Create a seperate one.

oSumAtrIX (changes requested):

app/src/main/java/app/revanced/integrations/patches/GeneralAdsPatch.java
0xrxL: There's a clean way to read litho component name outside general ads patch, without copy/paste general ads patch?

oSumAtrIX: Yes you create a new class and implement the same interface as `GeneralAdsPatch` does, make sure to add it to the same array `GeneralAdsPatch` is added to.

0xrxL: If I create a new class, I have no way to independently call it without copy-pasting the ads patch. So:

Can I extend it in the following way?

```public class Something extends GeneralAdsPatch```

Comment (0xrxL) [reply app/src/main/java/app/revanced/integrations/patches/GeneralAdsPatch.java]: > Instead you should do: `public final class Something extends Filter`

Oh...ok, I thought about this eventuality, but I wasn't sure if ```filter``` could work on other classes, without being directly called by general ads patch.

Btw...I'll push the new commit tomorrow.

**Review by Sculas — APPROVED** [review]

File: app/src/main/java/app/revanced/integrations/patches/LithoThemePatch.java [hunk 2]
```diff
@@ -20,12 +18,17 @@ public class LithoThemePatch {
...
+
     // Used by app.revanced.patches.youtube.layout.theme.patch.LithoThemePatch
     public static int applyLithoTheme(int originalValue) {
         var isDarkTheme = ThemeHelper.isDarkTheme();
         
-        if ((isDarkTheme && anyEquals(originalValue, DARKCONSTANTS)) || (!isDarkTheme  && anyEquals(originalValue, WHITECONSTANTS)))
-                return 0;
+        if ((isDarkTheme && anyEquals(originalValue, DARKCONSTANTS)) ||
+            (!isDarkTheme && anyEquals(originalValue, WHITECONSTANTS)))
+                return shortsCommentBox ? -16777215 : 0;
```
oSumAtrIX, line 30: Please create a final field for this magic constant the same way it is being done with the other constants.

oSumAtrIX (changes requested):

app/src/main/java/app/revanced/integrations/patches/LithoThemeComponentName.java
```diff
@@ -0,0 +1,11 @@
+package app.revanced.integrations.patches;
+
+import static app.revanced.integrations.patches.LithoThemePatch.shortsCommentBox;
+
+public final class LithoThemeComponentName extends Filter {
+    public boolean filter(final String path, final String identifier) {
+        shortsCommentBox = path.contains("comment_composer");
```
oSumAtrIX, line 7: Probably using a `BlockRule` object here would help with code consistency across extending classes of `Filter`. Consider using it here too.

oSumAtrIX (changes requested):

app/src/main/java/app/revanced/integrations/patches/LithoThemeComponentName.java
0xrxL: I don't think a blockrule it's needed here. It's simply a check after all, without the need for a return true but only false.

0xrxL: > Probably using a `BlockRule` object here would help with code consistency across extending classes of `Filter`. Consider using it here too.

Personal considerations aside: is this change strictly necessary?

oSumAtrIX: > I don't think a blockrule it's needed here. It's simply a check after all

A `BlockRule` object is an abstraction for that. I suggest it for code consistency and quality.

app/src/main/java/app/revanced/integrations/patches/LithoThemePatch.java
```diff
@@ -19,13 +17,22 @@ public class LithoThemePatch {
         -15790321, // comments chip background (new layout)
         -98492127 // video chapters list background
     };
+    private static final int solidBlack = -16777215;
+
+    private static final int trasparent = 0;
```
oSumAtrIX, line 22: Constants are written in big letters. Please apply the following changes:
```suggestion
    private static final int SOLID_BLACK = -16777215;
    private static final int TRANSPARENT = 0;
```

Also make sure to use `LIGHT` and `DARK` as the constant names for the fields above for consistency.

app/src/main/java/app/revanced/integrations/patches/LithoThemePatch.java
```diff
@@ -19,13 +17,22 @@ public class LithoThemePatch {
...
+
+    // boolean used to check if shorts comment box is currently visible
+    public static boolean shortsCommentBox = false;
```
oSumAtrIX, line 25: Usually getters and setters are used to access fields, but for that a seperate PR should be opened to fix this across the entire project consistently.

app/src/main/java/app/revanced/integrations/patches/LithoThemePatch.java
```diff
@@ -19,13 +17,22 @@ public class LithoThemePatch {
...
     // Used by app.revanced.patches.youtube.layout.theme.patch.LithoThemePatch
     public static int applyLithoTheme(int originalValue) {
         var isDarkTheme = ThemeHelper.isDarkTheme();
         
-        if ((isDarkTheme && anyEquals(originalValue, DARKCONSTANTS)) || (!isDarkTheme  && anyEquals(originalValue, WHITECONSTANTS)))
-                return 0;
+        if ((isDarkTheme && anyEquals(originalValue, DARKCONSTANTS)))
+            return shortsCommentBox ? solidBlack : trasparent;
+        if (!isDarkTheme && anyEquals(originalValue, WHITECONSTANTS)))
+            return trasparent;
```
oSumAtrIX, line 34: This gets rid of checking for `isDarkTheme` twice while retaining the code flow logic:

```suggestion
        if (isDarkTheme) {
            if (anyEquals(originalValue, DARKCONSTANTS))
                return shortsCommentBox ? solidBlack : trasparent;
        } else if (anyEquals(originalValue, WHITECONSTANTS))
            return trasparent;
```

oSumAtrIX (changes requested):

app/src/main/java/app/revanced/integrations/patches/LithoThemeComponentName.java
oSumAtrIX: You would add the same string `comment_composer` as a rule to block, then check for it and return the result to the method.

app/src/main/java/app/revanced/integrations/patches/LithoThemePatch.java
```diff
@@ -22,13 +22,25 @@ public class LithoThemePatch {
...
+            if (anyEquals(originalValue, DARKCONSTANTS))
+                return getColor("yt_black1");
+        } else if (anyEquals(originalValue, WHITECONSTANTS))
+            return getColor("yt_white1");
+
         return originalValue;
     }
 
+    private static int getColor(String colorName) {
+        return context.getColor(
```
oSumAtrIX, line 35: This should not be called this often. Instead store the color inside a field. I believe a getter also should be used instead accessing the field `context`. Please add a field for the return value of this method. Check if it is set every time the method is called, if not, set it, if it is set, return it.

oSumAtrIX (changes requested):

app/src/main/java/app/revanced/integrations/patches/LithoThemePatch.java
0xrxL: I don't know how to use getters and setters. I've never used them.

If you want, help me to fix this.

0xrxL: EDIT: Based on what I understand from your comment, the new commit should be enough.

oSumAtrIX: How are you sure context is set when the fields are set on static class initialization?

oSumAtrIX: Please do as previously suggested:

> Check if it is set every time the method is called, if not, set it, if it is set, return it.

0xrxL: Ok, thanks for clarifying this request further. I wasn't sure if I applied all the requests.

oSumAtrIX: Please check on [this](https://github.com/revanced/revanced-integrations/pull/239#discussion_r1054825073) as well.

0xrxL: Ok...so this is what I do:

1) Changed the static context getter to the method "getContext()", and created a dedicated field for the latter with a null check inside the getColor() method.

2) Moved the color fields setter to applyLithoTheme() method, and placed them inside their unique theme condition. In this way there will be a single field setter, based on the current theme.

3) Added field value checks, to avoid multiple getColor() method calls.

And I shouldn't have forgotten anything.

oSumAtrIX (approved):
I have left a final touch, please check if everything works correctly and we can get it merged.

0xrxL:
> I have left a final touch, please check if everything works correctly and we can get it merged.

It works without problems

Immagine 2022-12-22 005426

PR #240, opened 2022-12-19
refactor : AdRemoverAPI.HideViewWithLayout1dp in my opnion

hyunsu15:
it sames work maybe. because view.setLayoutParams's parameter is ViewGroup.Layout Params.

but i'm poor at andriod and it's inner logic.

ps.
I want to contribute other revanced projects. can i help you?

### DIFF
--- a/app/src/main/java/app/revanced/integrations/adremover/AdRemoverAPI.java
+++ b/app/src/main/java/app/revanced/integrations/adremover/AdRemoverAPI.java
@@ -3,11 +3,6 @@ package app.revanced.integrations.adremover;
 
 import android.view.View;
 import android.view.ViewGroup;
-import android.widget.FrameLayout;
-import android.widget.LinearLayout;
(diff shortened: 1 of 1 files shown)


oSumAtrIX:
Please explain what this PR does in detail. Why are you replacing everything?

hyunsu15:
i think that problem is miss casting.

maybe it can solve

```
 ViewGroup.LayoutParams layoutParam = ((ViewGroup) view).getLayoutParams();
            layoutParam.height=1;
            layoutParam.width=1;
            view.setLayoutParams(layoutParam);
``` 

reference:
https://stackoverflow.com/questions/21045198/android-view-viewgrouplayoutparams-cannot-be-cast-to-android-widget-abslistview

oSumAtrIX:
In that case it would not be able to hide all the other views which are not ViewGroups

PR #241, opened 2022-12-20
Fixed incorrect migration of renamed settings

LisoUseInAIKyrios:
Fixed oversight of the migration code.


oSumAtrIX (approved):
Good catch

PR #242, opened 2022-12-20
RemoveTodo & rename Request.method's name

hyunsu15:
Remove Todo as i can 
1. rename Request.method's name
2. remove private static method because it is same public method.

don' t use first commit [refactor : AdRemoverAPI.HideViewWithLayout1dp in my opnion](https://github.com/revanced/revanced-integrations/pull/242/commits/d094d246dd9ae3590f2fb1924de7081d4e5e6afe)

it applies already.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## [0.84.2](https://github.com/revanced/revanced-integrations/compare/v0.84.1...v0.84.2) (2022-12-20)
+
+
+### Bug Fixes
+
+* migrate old to new settings correctly ([#241](https://github.com/revanced/revanced-integrations/issues/241)) ([98ce2df](https://github.com/revanced/revanced-integrations/commit/98ce2df47dded785986dee3c64becf96956f5eb6))
+
 ## [0.84.1](https://github.com/revanced/revanced-integrations/compare/v0.84.0...v0.84.1) (2022-12-18)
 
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/app/revanced/integrations/sponsorblock/requests/SBRequester.java
```diff
@@ -178,7 +188,8 @@ public static void retrieveUserStats(PreferenceCategory category, Preference loa
     public static void setUsername(String username, EditTextPreference preference, Runnable toastRunnable) {
         new Thread(() -> {
             try {
-                HttpURLConnection connection = getConnectionFromRoute(SBRoutes.CHANGE_USERNAME, SettingsEnum.SB_UUID.getString(), username);
+                HttpURLConnection connection = getConnectionFromRoute(SBRoutes.CHANGE_USERNAME,
```
oSumAtrIX, line 191: Theres lots of unrelated refactoring changes, please separate refactoring as a different commit

oSumAtrIX (changes requested):

app/src/main/java/app/revanced/integrations/sponsorblock/requests/SBRequester.java
hyunsu15: problem is my auto template save. i will correct it.

PR #243, opened 2022-12-20
remove todo & rename Request's method name

hyunsu15:
past my problem corrects

 remove todo as i can
1. rename request's method name
2.  remove private static method because it;'s same public method

request other's methods are not touch
because parseJson is better than parseJsonAndClose.

other one is that it declared not disconnect but todo write disconnect it 's dismatching.


oSumAtrIX:
Does this require any change on the patches side?

oSumAtrIX:
I don't think this PR is necessary, it only appends `AndDisconnects` to all the methods which already have a comment on them saying the same. Thanks regardless.

PR #248, opened 2022-12-22
refactor DownloadButton

hyunsu15:
refactor : if statement optimizing

i will reduce this field

tell me field that will use

### DIFF
--- a/app/src/main/java/app/revanced/integrations/videoplayer/DownloadButton.java
+++ b/app/src/main/java/app/revanced/integrations/videoplayer/DownloadButton.java
@@ -109,16 +109,18 @@ public class DownloadButton {
 
         isShowing = z;
         ImageView imageView = _button.get();
-        if (_constraintLayout != null && imageView != null) {
-            if (z && isDownloadButtonEnabled) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/app/revanced/integrations/videoplayer/DownloadButton.java
```diff
@@ -109,16 +109,18 @@ public static void changeVisibility(boolean z) {
...
-                imageView.startAnimation(fadeIn);
-            } else if (imageView.getVisibility() == View.VISIBLE) {
-                LogHelper.printDebug(() -> "Fading out");
-                imageView.startAnimation(fadeOut);
-                imageView.setVisibility(View.GONE);
-            }
+        
+        if (_constraintLayout == null || imageView == null)
+            return ;
+        if (z && isDownloadButtonEnabled) {
```
oSumAtrIX, line 115:
```suggestion
            return;

        if (z && isDownloadButtonEnabled) {
```

oSumAtrIX (changes requested):

PR #251, opened 2022-12-26
feat(youtube): `double-back-to-close` patch

inotia00:
Make it possible to close the app by double-pressing the back button in the home feed

Also, it solves this issue https://github.com/revanced/revanced-patches/issues/615

### DIFF
--- a/app/src/main/java/app/revanced/integrations/patches/DouBleBackToClosePatch.java
+++ b/app/src/main/java/app/revanced/integrations/patches/DouBleBackToClosePatch.java
@@ -1,0 +1,66 @@
+package app.revanced.integrations.patches;
+
+import com.google.android.apps.youtube.app.watchwhile.WatchWhileActivity;
+
+import app.revanced.integrations.settings.SettingsEnum;
+import app.revanced.integrations.utils.LogHelper;
+
+public class DouBleBackToClosePatch {
+    /**
+     * Time between two back button presses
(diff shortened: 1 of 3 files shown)


oSumAtrIX:
Why should one want to press back two times when they can just press the home button to exit the app?

Kalpataru6:
People who use 3 button navigation need it..

inotia00:
there is no particular reason
It's just that when I first conceived this, I wanted to implement it as a feature that closes the app with a double tap.

If the home button can replace the role of the back button, there is no need for this PR.

Kalpataru6:
As you said there is no particular reason to do anything but functionality will be good.. I'm fine how it is now but my father always use back button and I can't for the love of God change his habit of back tapping to home button tapping.. If this implemented people like my father will be grateful.

oSumAtrIX:
The home button replaces the role of the back button. Your father would change his habit regardless of this PR because it only adds the feature to exit by swiping back twice.

Kalpataru6:
Oh ok.. But is there any way to solve the 3 button navigation 's back button issue?

oSumAtrIX:
Yes, by clicking the home button you can exit the app just like any other app.

0xrxL:
> Your father would change his habit

Apple style 😂

inotia00:
If so.. this PR is not necessary

PR #252, opened 2022-12-26
feat(youtube/return-youtube-dislike): better formatting and LTR support

LisoUseInAIKyrios:
- fixed RYD not working on devices that use right to left language layout (Arabic, Hebrew, etc):
https://freeimage.host/i/HIZBlNS

- restyled the additional like/dislike separator so it better matches the YouTube layout:
Current code: https://freeimage.host/i/HIQsNUv
These changes: https://freeimage.host/i/HIQvXa9

- small performance improvement:  onComponentCreated can be called multiple times for a single spannable text.  Check if the spannable was previously modified, and do not recreate if not needed.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1,0 +1,24 @@
+name: PR to main
+
+on:
+  push:
+    branches:
+      - dev
+  workflow_dispatch:
+
+env:
+  MESSAGE: merge branch \`${{ github.head_ref || github.ref_name }}\` to \`main\`
+
+jobs:
+  pull-request:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Open pull request
+        uses: repo-sync/pull-request@v2
+        with:
+          destination_branch: 'main'
+          pr_title: 'chore: ${{ env.MESSAGE }}'
+          pr_body: 'This pull request will ${{ env.MESSAGE }}.'
+          pr_draft: true
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -7,7 +7,7 @@ on:
   workflow_dispatch:
 
 env:
-  MESSAGE: merge branch \`${{ github.head_ref || github.ref_name }}\` to \`main\`
+  MESSAGE: merge branch `${{ github.head_ref || github.ref_name }}` to `main`
 
 jobs:
   pull-request:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -32,7 +32,7 @@ jobs:
     - name: Build with Gradle
       run: ./gradlew build --no-daemon
     - name: Setup semantic-release
-      run: npm install semantic-release @semantic-release/git @semantic-release/changelog gradle-semantic-release-plugin -D
+      run: npm install semantic-release @saithodev/semantic-release-backmerge @semantic-release/git @semantic-release/changelog gradle-semantic-release-plugin -D
     - name: Release
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,7 +29,7 @@ jobs:
       with:
         node-version: "latest"
     - name: Build with Gradle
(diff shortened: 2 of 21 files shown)


0xrxL:
>   These changes: https://freeimage.host/i/HIQvXa9

When I saw this image I thought about the possibility of adding an additional delimiter between the like symbol and its counter. In the following way:

HIQvXa9

Why that? Basically for a stupid aesthetic question:

1) It would give a sense of symmetry.
2) It would help temporarily cover the lack of a patch that could remove the delimiter before the dislike icon.

oSumAtrIX:
@0xrxL Unfortunately it is counterproductive to add more width to this component. The width added to the separator in #252 could also further increase the problem with the issue of the component cutting off the text.

0xrxL:
> @0xrxL Unfortunately it is counterproductive to add more width to this component. The width added to the separator in #252 could also further increase the problem with the issue of the component cutting off the text.

In this case the stock separator (the one before dislike icon) needs to be removed, also to improve the readability with the currently bugged UI layout.

LisoUseInAIKyrios:
> > @0xrxL Unfortunately it is counterproductive to add more width to this component. The width added to the separator in #252 could also further increase the problem with the issue of the component cutting off the text.
> 
> In this case the stock separator (the one before dislike icon) needs to be removed, also to improve the readability with the currently bugged UI layout.

the stock separator cannot be easily removed.

Adding another separator would not add that much horizontal space.

LisoUseInAIKyrios:
> @0xrxL Unfortunately it is counterproductive to add more width to this component. The width added to the separator in #252 could also further increase the problem with the issue of the component cutting off the text.

It still occasionally shows the like text as "120 | ...", but that already is happening without these changes.  I don't think this PR make the "..." occur more often than it currently does.

LisoUseInAIKyrios:
three separators
Quick test using 3 separators.  This would satisfy users desire to show the dislike value in a better place, as this looks more correct and less confusing.


The center separator could be any any unicode character (using a dot as an example):
two types of separators

LisoUseInAIKyrios:
> @0xrxL Unfortunately it is counterproductive to add more width to this component. The width added to the separator in #252 could also further increase the problem with the issue of the component cutting off the text.

I have not seen this PR display the likes as "120 | ..." (although I cannot verify it will never do this)
Maybe it is fixed by updating the likes label if needed (instead of changing it multiple times).  Although I'm not sure if the "..." is totally fixed.

0xrxL:
> Quick test using existing layout, two separators, and 3 separators. current appearance two separators three horizontal separators round separator square separator
> 
> 3 separators would partially satisfy users desire to show the dislike value in a better place (issue [1140](https://github.com/revanced/revanced-patches/issues/1140)), as this looks more correct and less confusing.
> 
> The center separator could be any unicode character (shown above with a dot, and a square) I think the dot or square layout looks pretty good.

The solution with the center dash (fourth picture) seems good.

0xrxL:
> > @0xrxL Unfortunately it is counterproductive to add more width to this component. The width added to the separator in #252 could also further increase the problem with the issue of the component cutting off the text.
> 
> I have not seen this PR display the likes as "89K | ..."

Try to tap on save to playlist button, and check if something happens.

LisoUseInAIKyrios:
> Try to tap on save to playlist button, and check if something happens.

The likes layout changes to "89k | ..."
So it's not entirely fixed.

0xrxL:
> > Try to tap on save to playlist button, and check if something happens.
> 
> The likes layout changes to "89k | ..." So it's not entirely fixed.

This is a stock bug.

LisoUseInAIKyrios:
It now shows 3 separators (actual screen shot):

3 separators

LisoUseInAIKyrios:
It now shows 3 separators (actual screen shot):

3 separators

0xrxL:
@oSumAtrIX I think this PR is good enough to be reviewed (and eventually merged).

Adding the extra delimiter does not cause the buttons name clipping bug (the one with 3 dots)...obviously not until Save button will not be clicked.

But in any case, even in the absence of the extra delimiter, the button names would still be cut off by the bug.

oSumAtrIX:
New change:

image

oSumAtrIX:
Its not perfect yet. I assume it will also depend on the font:

image

PR #255, opened 2022-12-27
refactor :seperate DownloadButtonAnimation

hyunsu15:
i work that separateDownloadButtonAnimation

because DownloadButton has many field and mutable field

so, i seperate DownloadButtonAnimation

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1,0 +1,24 @@
+name: PR to main
+
+on:
+  push:
+    branches:
+      - dev
+  workflow_dispatch:
+
+env:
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,205 @@
+# 1.0.0 (2022-12-27)
```
oSumAtrIX, line 1: Please do not commit to this file, it is autogenerated.

oSumAtrIX (changes requested):
What is the reason for abstracting the animation-related methods to a `DownloadButtonAnimation` class? Instead, an abstract class should be created which holds everything related to buttons so when we need to add new buttons, duplicate code can be avoided.

PR #256, opened 2022-12-28
refactor : seperate DownloadButtonAnimation

hyunsu15:
oSumAtrIX's opinion

What is the reason for abstracting the animation-related methods to a DownloadButtonAnimation class? Instead, an abstract class should be created which holds everything related to buttons so when we need to add new buttons, duplicate code can be avoided.

then my opion

your opinion is also correct.
but from Software Engineering point of view, past code seems to lack cohesion.
good code for maintenance requires higher cohesion
So I gave this opinion. I think I should make it every time if necessary.

commit mistake is poor at github
I'll try to fix it as soon as possible.


oSumAtrIX (commented):
The new class does not really make sense from an object-oriented perspective. As previously suggested and also mentioned in the PR description, OOP should be respected in that regard. Thank you for the PR anyways! If you want to make such changes in the future, I recommend opening an issue and discussing your upcoming changes before investing your time in them.

hyunsu15:
Do you know  article or links or posts or examples that related that part?

I want to know it.

PR #261, opened 2022-12-30
fix(youtube/general-ads-patch): improved ignore list

0xrxL:
As I explained in a previous occasion, some of the current ignore list component names are totally useless. Today I tried to remove them from the list and are still visible.

This is the screenshot of the final result:

Immagine 2022-12-30 123620

### DIFF
--- a/app/src/main/java/app/revanced/integrations/patches/GeneralAdsPatch.java
+++ b/app/src/main/java/app/revanced/integrations/patches/GeneralAdsPatch.java
@@ -13,9 +13,7 @@ public final class GeneralAdsPatch extends Filter {
             "related_video_with_context",
             "comment_thread", // skip blocking anything in the comments
             "|comment.", // skip blocking anything in the comments replies
-            "download_",
             "library_recent_shelf",
-            "playlist_add_to_option_wrapper" // do not block on "add to playlist" flyout menu
     };
 
     private final BlockRule custom = new CustomBlockRule(


0xrxL:
...and both are still visible with all layout settings (of ReVanced settings) enabled

Immagine 2022-12-30 124348

oSumAtrIX:
Please explain, why removing them does not cause any issues. Respond with the new component names.

0xrxL:
As you know, the contains() method verifies that an entire word or a series of letters is contained in a string. So, if a filter name is (totally or partially) contained in two component names for example, both will be blocked, also if you want just to block one of them.

Same thing happened before before ```general ads patch``` refactor, with  names: ```download_``` and ```playlist_add_wrapper```

0xrxL:
The following screenshot shows a list of ads to block, prior to your refactor:

Screenshot_20221230_173652_Samsung Internet Beta

As you can already guess, contains() method return a true value because the string ```ad_``` is contained in the name ```download_``` and also in a name of an ad component. So both components were blocked.

0xrxL:
This instead is the current ads filtering list (the one after your refactor):

Screenshot_20221230_174348_Samsung Internet Beta

As you can see, ```ad_``` and ```_ad``` names not exists anymore on the ads filtering list. So, there's no point to keep ```download_``` and ```playlist_add_wrapper``` on the ignore list.

0xrxL:
To sum up:

With your refactor of the ad blocking list, you inadvertently fixed a problem that required adding two component names to the ignore list to prevents their blocking.

oSumAtrIX:
Good explanation. Next time, please add such explanations in a PR when it is entirely clear.

0xrxL:
> Good explanation. Next time, please add such explanations in a PR when it is entirely clear.

I will remember to do it.

PR #263, opened 2022-12-30
feat(youtube): `copy-video-url` patch

aliernfrog:
Adds required code for copy video & copy timestamp URL buttons in video player. Probably needs improvements.
https://github.com/revanced/revanced-patches/pull/1402


oSumAtrIX (changes requested):
Overall, this code is a duplicate of existing code. There are already methods to copy the clipboard or classes dedicated to showing a button such as the download button which is why the current code has to be properly refactored to get rid of the duplicate code.

aliernfrog:
I couldn't find any clipboard methods in current code.
To resolve the duplicated code issue, i thought about creating a new `BottomControlButton` class and using it for all 3 buttons, would that be fine?

oSumAtrIX:
You are right, I was likely remembering something else instead of clipboard methods. Either way, clipboard methods are utility methods and can be put into the ReVanced utility class if you want to and see a reason to. A `BottomControlButton` class sounds good.

aliernfrog:
Made following changes:
- Moved clipboard method to `ReVancedUtils`
- Added video id to VideoInformation and made use of that in downloads and copy video url patch
- Added `BottomControlButton` class to avoid duplicate code

app/src/main/java/app/revanced/integrations/videoplayer/CopyVideoUrlTimestampButton.java
```diff
@@ -0,0 +1,21 @@
+package app.revanced.integrations.videoplayer;
+
+import app.revanced.integrations.patches.CopyVideoUrlPatch;
+import app.revanced.integrations.settings.SettingsEnum;
+
+public class CopyVideoUrlTimestampButton {
+    static BottomControlButton _button;
```
oSumAtrIX, line 7: Is there a specific reason the button classes do not extend on it?

oSumAtrIX:
Looks good so far, I believe after a final review we can get this merged.

app/src/main/java/app/revanced/integrations/videoplayer/CopyVideoUrlTimestampButton.java
aliernfrog: I thought about that but couldn't figure out how it would work.

oSumAtrIX: I'll give it a look then.

oSumAtrIX: I have pushed some changes, please review & make a final test so we can merge it.

aliernfrog: Also pushed a change to remove video time hook, will test in a bit.

aliernfrog: Added null checks to button instances, since `changeVisibility` probably gets called before the button gets initialized. After doing this change, everything works as intended.

oSumAtrIX (approved):

PR #264, opened 2022-12-31
fix(youtube/general-ads): restore swipe back to exit gesture

0xrxL:


app/src/main/java/com/google/android/apps/youtube/app/watchwhile/WatchWhileActivity.java
```diff
@@ -0,0 +1,7 @@
+package com.google.android.apps.youtube.app.watchwhile;
```
oSumAtrIX, line 1: This is the wrong package for dummy classes. Move this class to `https://github.com/revanced/revanced-integrations/tree/main/dummy/src/main/java/com/google/android/apps/youtube/app`.

app/src/main/java/app/revanced/integrations/patches/BackToExitPatch.java
```diff
@@ -0,0 +1,54 @@
+package app.revanced.integrations.patches;
+
+import com.google.android.apps.youtube.app.watchwhile.WatchWhileActivity;
+import app.revanced.integrations.utils.LogHelper;
+
+public class BackToExitPatch {
+    private static long backPressedCount = 0;
+
+    /**
+     * State whether scroll position reaches the top
```
oSumAtrIX, line 10:
```suggestion
     * State whether the scroll position reaches the top.
```

app/src/main/java/app/revanced/integrations/patches/BackToExitPatch.java
```diff
@@ -0,0 +1,54 @@
...
+     */
+    public static boolean isHomeScrolledTop = false;
+
+    /**
+     * Detect event when back button is pressed
+     *
+     * @param activity is used when closing the app
+     */
```
oSumAtrIX, line 18:
```suggestion
    /**
     * Handle the event of clicking the back button.
     *
     * @param activity The activity, the app is launched with to finish.
     */
```

app/src/main/java/app/revanced/integrations/patches/BackToExitPatch.java
```diff
@@ -0,0 +1,54 @@
...
+    public static void exitOnBackPressed(WatchWhileActivity activity) {
+        // Check scroll position reaches the top in home feed
```
oSumAtrIX, line 20:
```suggestion
```

app/src/main/java/app/revanced/integrations/patches/BackToExitPatch.java
```diff
@@ -0,0 +1,54 @@
...
+        // Close the current activity once counter reached 1
+        if (backPressedCount <= 0) {
+            LogHelper.printDebug(() -> "BackPressed detected");
+
+            backPressedCount++;
+        } else {
+            LogHelper.printDebug(() -> "Activity is finished");
+
+            activity.finish();
+        }
```
oSumAtrIX, line 32: This makes no sense. Why is this check required? The activity should be finished when `isHomeScrolledTop` is true immediately, when `exitOnBackPressed` is called.

app/src/main/java/app/revanced/integrations/patches/BackToExitPatch.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    /**
+     * Detect event when ScrollView is created by RecyclerView
+     */
```
oSumAtrIX, line 37:
```suggestion
    /**
     * Handle the event, when it is being scrolled.
     */
```

app/src/main/java/app/revanced/integrations/patches/BackToExitPatch.java
```diff
@@ -0,0 +1,54 @@
...
+        isHomeScrolledTop = false;
+
+        backPressedCount = 0;
+
+        LogHelper.printDebug(() -> "ScrollView is created");
+    }
+
+    /**
+     * Detect event when the scroll position reaches the top by the back button
+     */
```
oSumAtrIX, line 48:
```suggestion
    /**
     * Handle the event of stopping to scroll.
     */
```

app/src/main/java/app/revanced/integrations/patches/BackToExitPatch.java
```diff
@@ -0,0 +1,54 @@
...
+    public static void onStopScrollView() {
+        isHomeScrolledTop = true;
+
+        LogHelper.printDebug(() -> "ScrollView is stopped");
```
oSumAtrIX, line 52:
```suggestion
        LogHelper.printDebug(() -> "Stopped scrolling");
```

app/src/main/java/app/revanced/integrations/patches/BackToExitPatch.java
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * Detect event when ScrollView is created by RecyclerView
+     */
+    public static void onStartScrollView() {
+        isHomeScrolledTop = false;
+
+        backPressedCount = 0;
+
+        LogHelper.printDebug(() -> "ScrollView is created");
```
oSumAtrIX, line 43:
```suggestion
        LogHelper.printDebug(() -> "Started scrolling");
```

oSumAtrIX (changes requested):

app/src/main/java/app/revanced/integrations/patches/BackToExitPatch.java
```diff
@@ -0,0 +1,41 @@
+package app.revanced.integrations.patches;
+
+import com.google.android.apps.youtube.app.watchwhile.WatchWhileActivity;
+import app.revanced.integrations.utils.LogHelper;
+
+public class BackToExitPatch {
```
oSumAtrIX, line 6: This class name should be the same as the counterpart patch class.

app/src/main/java/app/revanced/integrations/patches/BackToExitPatch.java
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * Handle the event of clicking the back button.
+     *
+     * @param activity The activity, the app is launched with to finish.
+     */
+    public static void exitOnBackPressed(WatchWhileActivity activity) {
+        if (!isHomeScrolledTop) return;
+
+        // Close the current activity
```
oSumAtrIX, line 20:
```suggestion
```

app/src/main/java/app/revanced/integrations/patches/BackToExitPatch.java
```diff
@@ -0,0 +1,41 @@
...
+        activity.finish();
+    }
```
oSumAtrIX, line 22:
```suggestion
        
        LogHelper.printDebug(() -> "Closed activity");
    }
```

oSumAtrIX (changes requested):

app/src/main/java/app/revanced/integrations/patches/BackToExitPatch.java
```diff
@@ -0,0 +1,42 @@
...
+    public static void onStartScrollView() {
+        isHomeScrolledTop = false;
+
+        LogHelper.printDebug(() -> "Started scrolling");
+    }
+
+    /**
+     * Handle the event of stopping to scroll.
+     */
+    public static void onStopScrollView() {
```
oSumAtrIX, line 37: From what the code and comments convey, is that when scrolling stops, the next click on the back button will close the app. This should not be the case. Instead it should only close, when - as the variable name `isHomeScrolledTop` suggests - the feed is scrolled to top.

0xrxL: So what you suggest? To change the variables/methods names?

0xrxL: Anyway...the current code close the activity immediately after the list is entirely scrolled to the up (until the first video).

So should be good.

oSumAtrIX: > So what you suggest? To change the variables/methods names?

If the behavior works as intended, the comment and method name should be adjusted to reflect what they really do.

PR #265, opened 2022-12-31
fix(youtube/return-youtube-dislike): use manufacturer specific span styles

LisoUseInAIKyrios:
The OnePlus changes pushed my Samsung and Google phones out of wack.

No set of ratios will be perfect for all devices, all we can get is good enough for most devices

Can you see if these changes looks ok on your device?

Here's what it this PR looks like on my end:

Google Pixel (with this PR):
gogle pixel

Samsung (with this PR):
samsung

### DIFF
--- a/app/src/main/java/app/revanced/integrations/returnyoutubedislike/ReturnYouTubeDislike.java
+++ b/app/src/main/java/app/revanced/integrations/returnyoutubedislike/ReturnYouTubeDislike.java
@@ -326,7 +326,8 @@ public class ReturnYouTubeDislike {
                 }
                 // shift everything up, to compensate for the vertical movement caused by the font change below
                 // each section needs it's own Relative span, otherwise alignment is wrong
-                addSpanStyling(leftSeparatorSpan, new RelativeVerticalOffsetSpan(segmentedVerticalShiftRatio));
+                addSpanStyling(leftSeparatorSpan, new RelativeVerticalOffsetSpan(segmentedLeftSeparatorVerticalShiftRatio));
+
                 addSpanStyling(likesSpan, new RelativeVerticalOffsetSpan(segmentedVerticalShiftRatio));
                 addSpanStyling(middleSeparatorSpan, new RelativeVerticalOffsetSpan(segmentedVerticalShiftRatio));
                 addSpanStyling(dislikeSpan, new RelativeVerticalOffsetSpan(segmentedVerticalShiftRatio));
@@ -352,6 +353,7 @@ public class ReturnYouTubeDislike {
 
     private static boolean segmentedValuesSet = false;
     private static float segmentedVerticalShiftRatio;
+    private static float segmentedLeftSeparatorVerticalShiftRatio;
     private static float segmentedLeftSeparatorFontRatio;
     private static float segmentedLeftSeparatorHorizontalScaleRatio;
 
@@ -380,7 +382,7 @@ public class ReturnYouTubeDislike {
                 configManufacturer = "Google";
                 configSdk = 33;
                 // tested on Android 10 thru 13, and works well for all
-                segmentedVerticalShiftRatio = -0.18f; // move separators and like/dislike up by 18%
+                segmentedLeftSeparatorVerticalShiftRatio = segmentedVerticalShiftRatio = -0.18f; // move separators and like/dislike up by 18%
                 segmentedLeftSeparatorFontRatio = 1.8f;  // increase left separator size by 80%
                 segmentedLeftSeparatorHorizontalScaleRatio = 0.65f; // horizontally compress left separator by 35%
                 break;
@@ -388,7 +390,7 @@ public class ReturnYouTubeDislike {
                 configManufacturer = "samsung";
                 configSdk = 33;
                 // tested on S22
-                segmentedVerticalShiftRatio = -0.19f;
+                segmentedLeftSeparatorVerticalShiftRatio = segmentedVerticalShiftRatio = -0.19f;
                 segmentedLeftSeparatorFontRatio = 1.5f;
                 segmentedLeftSeparatorHorizontalScaleRatio = 0.7f;
                 break;
--- a/app/src/main/java/app/revanced/integrations/returnyoutubedislike/ReturnYouTubeDislike.java
+++ b/app/src/main/java/app/revanced/integrations/returnyoutubedislike/ReturnYouTubeDislike.java
@@ -331,7 +331,7 @@ public class ReturnYouTubeDislike {
                 final float leftSeparatorFontRatio = 1.6f;  // increase height by 60%
                 final float leftSeparatorHorizontalScaleRatio = 0.6f; // compress horizontally by 40%
 
-                // shift the left separator up by a smaller amount, to visually align it after changing the size
+                // shift everything up, to compensate for the vertical movement caused by the font change below
                 addSpanStyling(leftSeparatorSpan, new RelativeVerticalOffsetSpan(verticalShiftRatio));
                 addSpanStyling(likesSpan, new RelativeVerticalOffsetSpan(verticalShiftRatio));
                 addSpanStyling(middleSeparatorSpan, new RelativeVerticalOffsetSpan(verticalShiftRatio));
--- a/app/src/main/java/app/revanced/integrations/returnyoutubedislike/ReturnYouTubeDislike.java
+++ b/app/src/main/java/app/revanced/integrations/returnyoutubedislike/ReturnYouTubeDislike.java
@@ -345,7 +345,7 @@ public class ReturnYouTubeDislike {
                 addSpanStyling(middleSeparatorSpan, new RelativeVerticalOffsetSpan(verticalShiftRatio));
                 addSpanStyling(dislikeSpan, new RelativeVerticalOffsetSpan(verticalShiftRatio));
 
-                // important: must set font size after vertical offset (otherwise alignment gets off)
+                // important: must add size scaling after vertical offset (otherwise alignment gets off)
                 addSpanStyling(leftSeparatorSpan, new RelativeSizeSpan(leftSeparatorFontRatio));
                 addSpanStyling(leftSeparatorSpan, new ScaleXSpan(leftSeparatorHorizontalScaleRatio));
                 // middle separator does not need resizing
--- a/app/src/main/java/app/revanced/integrations/returnyoutubedislike/ReturnYouTubeDislike.java
+++ b/app/src/main/java/app/revanced/integrations/returnyoutubedislike/ReturnYouTubeDislike.java
@@ -11,7 +11,6 @@ import android.text.SpannableStringBuilder;
 import android.text.TextPaint;
 import android.text.style.CharacterStyle;
 import android.text.style.ForegroundColorSpan;
-import android.text.style.MetricAffectingSpan;
 import android.text.style.RelativeSizeSpan;
 import android.text.style.ScaleXSpan;
 import android.util.DisplayMetrics;
--- a/app/src/main/java/app/revanced/integrations/returnyoutubedislike/ReturnYouTubeDislike.java
+++ b/app/src/main/java/app/revanced/integrations/returnyoutubedislike/ReturnYouTubeDislike.java
@@ -321,7 +321,8 @@ public class ReturnYouTubeDislike {
                 // but it's good enough and still visually better than not doing this scaling/shifting
                 //
                 // important: any changes made here should be tested on a few different devices.
-                // Getting these values perfect on one device, may looks wrong on another device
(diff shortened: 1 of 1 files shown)


oSumAtrIX:
I believe adding settings to ReVanced which allow to individually set those parameters would be a good idea instead of settling down with a single configuration. In that case this PR can be adjusted to that.

LisoUseInAIKyrios:
Another PR can add settings for users to wiggle into place, but this PR is about the default alignment.

I think the main issue is the font scale is set before the vertical offset, which does not uniformly apply to different devices.  I found the relative font size must be set after the vertical offset.

Here is what the current dev branch looks like:

Google:
google_pixel_dev_branch

Samsung:
samsung_dev_branch

oSumAtrIX:
I believe making it align properly for Google and Samsung devices is better for a larger set of users.

On OnePlus it looks like this:

Screenshot_2022-12-31-08-19-40-56_64ef5fc2000c1caa954c114bb372e1d5

oSumAtrIX:
Can you send a picture what the PR looks like for you?

LisoUseInAIKyrios:
This PR looks like screenshots in the [first comment](https://github.com/revanced/revanced-integrations/pull/265#issue-1515037433)

LisoUseInAIKyrios:
> I believe making it align properly for Google and Samsung devices is better for a larger set of users.
> 
> On OnePlus it looks like this:
> 
> Screenshot_2022-12-31-08-19-40-56_64ef5fc2000c1caa954c114bb372e1d5

Is this a screenshot of this PR?

LisoUseInAIKyrios:
> I believe making it align properly for Google and Samsung devices is better for a larger set of users.
> 
> On OnePlus it looks like this:
> 
> Screenshot_2022-12-31-08-19-40-56_64ef5fc2000c1caa954c114bb372e1d5

Is this a screenshot of this PR?  If so I think it looks ok.  I increase the horizontal separator compression a bit, which might fix the left separator looking too thick.

LisoUseInAIKyrios:
Try this PR again, and adjust 'leftSeparatorHorizontalScaleRatio' if needed.

Otherwise I think it looks ok for these 3 major brands.

LisoUseInAIKyrios:
Actually it might be best to add a switch statement and adjust based on device manufacturer.  I will try this shortly.

LisoUseInAIKyrios:
Ok its not too complicated, and if anyone cares they can add additional devices.

It's now pixel perfect on Samsung and Google phones running Android 13:

Google Pixel

Samsung

app/src/main/java/app/revanced/integrations/returnyoutubedislike/ReturnYouTubeDislike.java
```diff
@@ -331,24 +328,15 @@ public void updateDrawState(TextPaint tp) {
...
-                addSpanStyling(leftSeparatorSpan, new ScaleXSpan(horizontalStretchRatio));
-
-                // shift the left separator up by a smaller amount, to visually align it after changing the size
-                addSpanStyling(leftSeparatorSpan, new RelativeVerticalOffsetSpan(verticalLeftSeparatorShiftRatio));
-                addSpanStyling(likesSpan, new RelativeVerticalOffsetSpan(verticalShiftRatio));
-                addSpanStyling(middleSeparatorSpan, new RelativeVerticalOffsetSpan(verticalShiftRatio));
-                addSpanStyling(dislikeSpan, new RelativeVerticalOffsetSpan(verticalShiftRatio));
-
+                // shift everything up, to compensate for the vertical movement caused by the font change below
+                addSpanStyling(leftSeparatorSpan, new RelativeVerticalOffsetSpan(segmentedVerticalShiftRatio));
```
oSumAtrIX, line 332: Is it really necessary to instantiate the class every time `updateDislike` is called?

oSumAtrIX (changes requested):

app/src/main/java/app/revanced/integrations/returnyoutubedislike/ReturnYouTubeDislike.java
LisoUseInAIKyrios: Instantiate the style spans?  Maybe they could be reused, not sure though.  But they are not heavy weight objects or expensive to create.

LisoUseInAIKyrios: I tried using a single instance of RelativeVericalOffsetSpan across all of the different spans (like, dislike, separators), but the alignment was wrong.  Using individual offset spans fixed that.

LisoUseInAIKyrios: Just tried again to confirm, and each span needs it's own instance.  (otherwise the sections don't like up vertically)

LisoUseInAIKyrios (commented):
It looks like these colors might be resource 'yt_separator_dark', and 'yt_separator_light'?

But I can't figure out how to find the colors associated with those resources.

LisoUseInAIKyrios:
Ok I was able to load them using some borrowed code from LithoThemePatch#getColor

Although those do not appear to be exactly the right colors, they do look close.

oSumAtrIX:
The colors can be found in the colors.xml file when decoding the resources. Although Litho and the likes work a bit differently, selecting the colors by eye should suffice.

LisoUseInAIKyrios:
> The colors can be found in the colors.xml file when decoding the resources. Although Litho and the likes work a bit differently, selecting the colors by eye should suffice.

The light color can be picked by eye, but the dark color is more difficult because of the transparency.

When in dark mode, some of the video colors are cast into area surrounding the video.  The dark color has a transparency, which causes it to show up differently depending on what video is playing.  Very difficult to figure out the transparency level using photoshop.

oSumAtrIX:
The transparency value can be calculated by picking the hex of the bleeding color and the hex of the separator with the bleeding color below. Then you diff them and have the alpha for the transparency.

LisoUseInAIKyrios:
I tried calculating the transparency, but I could not find the correct color and transparency.

The issue I ran into is the gray oval button container is also transparent, which means the video color bleeds thru the button container and then bleeds thru the separator as well.  So the separator appearance is a combination of  3 colors (video bleed, button container, and the separators own color).

If you have time, then you can give it a go to figure out the correct transparency.  Otherwise I will use trial and error and find a transparency and color value that gets close but is not perfect.

LisoUseInAIKyrios:
I found a dark mode color that looks ok
red video

dark video

0xrxL:
> I found a dark mode color that looks ok red video
> 
> dark video

Great! But why you didn't get it with the method ```context.getColor()``` instead?

At least you will not have to adjust it manually if there is a need.

LisoUseInAIKyrios:
I found that color by hand.

I did not find an exact color match in the resources.  I think the stock separator is drawn using an tiled image.  Or possibly it's using a resource color but with an additional unknown transparency.

So for now, ReVanced will use a hand selected color.

0xrxL:
> I found that color by hand.
> 
> I did not find an exact color match in the resources. I think the stock separator is drawn using an tiled image. Or possibly it's using a resource color but with an additional unknown transparency.
> 
> So for now, ReVanced will use a hand selected color.

You find a color for both white/dark theme, right?

LisoUseInAIKyrios:
White theme color was chosen by hand as well, and it was already a 100% match since there's no color bleed to deal with.

0xrxL:
> White theme color was chosen by hand as well, and it was already a 100% match since there's no color bleed to deal with.

Ok, at the least there will be no problems of colors with one of two themes.

app/src/main/java/app/revanced/integrations/returnyoutubedislike/ReturnYouTubeDislike.java
```diff
@@ -255,7 +255,7 @@ private static boolean updateDislike(AtomicReference<Object> textRef, boolean is
             }
             replacementSpannable = newSpannableWithDislikes(oldSpannable, voteData);
         } else {
-            String leftSegmentedSeparatorString = ReVancedUtils.isRightToLeftTextLayout() ? "\u200F|  " : "|  ";
+            String leftSegmentedSeparatorString = ReVancedUtils.isRightToLeftTextLayout() ? "\u200F|   " : "|   ";
```
oSumAtrIX, line 258: Why has a space been added?

app/src/main/java/app/revanced/integrations/returnyoutubedislike/ReturnYouTubeDislike.java
```diff
@@ -365,6 +350,81 @@ public void updateDrawState(TextPaint tp) {
...
+        final int deviceHeight = metric.heightPixels;
+        LogHelper.printDebug(() -> "Device manufacturer: '" + deviceManufacturer + "' SDK: " + deviceSdkVersion
+                + " scaledDensity: " + deviceScaledDensity + " densityDpi: " + deviceDensityDpi
+                + " width: " + deviceWidth + " height: " + deviceHeight);
+
+        // device used to determine the configuration below
+        final String configManufacturer; // very important
+        final int configSdk; // important
+        final float configScaledDensityDpi; // important only if using non default system font size
+        final int configDensityDpi, configWidth, configHeight; // rarely important
```
oSumAtrIX, line 381: These are not used anywhere except for this method. What is the purpose of calling `setSegmentedAdjustmentValues()` if it does not change anything?

oSumAtrIX (changes requested):

oSumAtrIX:
I believe, instead of adding multiple configurations for devices, a default config would suffice in addition with the ability to change the parameters such as offsets for height and size or color adjustments for the span.

app/src/main/java/app/revanced/integrations/returnyoutubedislike/ReturnYouTubeDislike.java
LisoUseInAIKyrios: 3 spaces looks a little better on the devices I used.  Although 2 looks ok as well I guess.  I'll change it back to 2 spaces.

oSumAtrIX: You can change the width of a span that contains the space and append it.

app/src/main/java/app/revanced/integrations/returnyoutubedislike/ReturnYouTubeDislike.java
LisoUseInAIKyrios: The config variables are for documentation, to show what device was used for the variables segmentedVerticalShiftRatio, segmentedLeftSeparatorFontRatio, segmentedLeftSeparatorHorizontalScaleRatio.

The config variable documentation could be left out, but then it's a mystery what device was used to come up with the settings.

LisoUseInAIKyrios:
> I believe, instead of adding multiple configurations for devices, a default config would suffice in addition with the ability to change the parameters such as offsets for height and size or color adjustments for the span.

The google config is the default, and seems to work ok on most devices.  The setSegmentedAdjustmentValues method allows adding whatever configs anyone wants, and it can be as coarse or fine grained as needed.

I think providing in app user settings to tweak the layout would work, but I think 99% of users will ignore it

LisoUseInAIKyrios:
> Here is my test on Samsung Galaxy A20, it look like not work properly image

Is that using the latest changes of this PR?  Is your device Android 13?

LisoUseInAIKyrios:
> Here is my test on Samsung Galaxy A20, it look like not work properly
Screenshot_20230102_110116
Here's this PR on S22 Android 13

app/src/main/java/app/revanced/integrations/returnyoutubedislike/ReturnYouTubeDislike.java
LisoUseInAIKyrios: To simplify this, I removed all the config options except the manufacturer and android build version.

LisoUseInAIKyrios:
> Galaxy A20

My mistake, I went on the first Google search result.

Is your screenshot from this PR?  Or from the current release?

LisoUseInAIKyrios:
I'm not a bro.  Can you post a screen shot of the latest release?  I am unable to reproduce when using Android 11.

LisoUseInAIKyrios:
The latest ReVanced release looks like this
https://github.com/revanced/revanced-integrations/pull/265#issuecomment-1368177429

LisoUseInAIKyrios:
> Yes, I meant that. And now I have tested on the latest version and have the same error like above too !

You should open a bug report, as the problem exists outside this PR.  In the bug report include screenshots of both regular and dark mode on your device.

oSumAtrIX (approved):

PR #281, opened 2023-01-07
RYD: better color match of existing dark mode separator color

LisoUseInAIKyrios:
I got a better match of the separator color.  This is about the best it's going to get.

black background video
red background video
green background video


LisoUseInAIKyrios:
"12 hours of the color (whatever)!"  is not popular among viewers.

What did they expect?  They got exactly what the video says.

oSumAtrIX (approved):

PR #284, opened 2023-01-08
feat(YouTube): ReturnYouTubeDislike compact like button layout

LisoUseInAIKyrios:
integration changes for https://github.com/revanced/revanced-patches/pull/1454

### DIFF
--- a/app/src/main/java/app/revanced/integrations/returnyoutubedislike/ReturnYouTubeDislike.java
+++ b/app/src/main/java/app/revanced/integrations/returnyoutubedislike/ReturnYouTubeDislike.java
@@ -254,9 +254,11 @@ public class ReturnYouTubeDislike {
             }
             replacementSpannable = newSpannableWithDislikes(oldSpannable, voteData);
         } else {
-            String leftSegmentedSeparatorString = ReVancedUtils.isRightToLeftTextLayout() ? "\u200F|  " : "|  ";
+            final boolean useCompactLayout = SettingsEnum.RYD_USE_COMPACT_LAYOUT.getBoolean();
+            // if compact layout, use a "half space" character
+            String middleSegmentedSeparatorString = useCompactLayout ? "\u2009 • \u2009" : "  •  ";
 
-            if (oldLikesString.contains(leftSegmentedSeparatorString)) {
+            if (oldLikesString.contains(middleSegmentedSeparatorString)) {
                 return false; // dislikes was previously added
             }
 
@@ -287,14 +289,11 @@ public class ReturnYouTubeDislike {
             } else {
                 Spannable likesSpan = newSpanUsingStylingOfAnotherSpan(oldSpannable, oldLikesString);
 
-                // left and middle separator
-                String middleSegmentedSeparatorString = "  •  ";
-                Spannable leftSeparatorSpan = newSpanUsingStylingOfAnotherSpan(oldSpannable, leftSegmentedSeparatorString);
+                // middle separator
                 Spannable middleSeparatorSpan = newSpanUsingStylingOfAnotherSpan(oldSpannable, middleSegmentedSeparatorString);
                 final int separatorColor = ThemeHelper.isDarkTheme()
                         ? 0x29AAAAAA  // transparent dark gray
                         : 0xFFD9D9D9; // light gray
-                addSpanStyling(leftSeparatorSpan, new ForegroundColorSpan(separatorColor));
                 addSpanStyling(middleSeparatorSpan, new ForegroundColorSpan(separatorColor));
                 CharacterStyle noAntiAliasingStyle = new CharacterStyle() {
                     @Override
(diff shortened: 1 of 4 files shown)


app/src/main/java/app/revanced/integrations/returnyoutubedislike/ReturnYouTubeDislike.java
```diff
@@ -254,9 +254,11 @@ private static boolean updateDislike(AtomicReference<Object> textRef, boolean is
             }
             replacementSpannable = newSpannableWithDislikes(oldSpannable, voteData);
         } else {
-            String leftSegmentedSeparatorString = ReVancedUtils.isRightToLeftTextLayout() ? "\u200F|  " : "|  ";
+            final boolean useCompactLayout = SettingsEnum.RYD_USE_COMPACT_LAYOUT.getBoolean();
+            // if compact layout, use a "half space" character
+            String middleSegmentedSeparatorString = useCompactLayout ? "\u2009 • \u2009" : "  •  ";
```
oSumAtrIX, line 259: Does this really make a difference?

image

image

app/src/main/java/app/revanced/integrations/returnyoutubedislike/ReturnYouTubeDislike.java
```diff
@@ -254,9 +254,11 @@ private static boolean updateDislike(AtomicReference<Object> textRef, boolean is
...
 
-            if (oldLikesString.contains(leftSegmentedSeparatorString)) {
+            if (oldLikesString.contains(middleSegmentedSeparatorString)) {
```
oSumAtrIX, line 261: Currently, this works, but it's a "hacky" way. You can depend on the `video-information` patch and check if the video has changed and determine this way if it is necessary to update the string. Not a necessary change and unrelated to this PR but something to consider in the future.

app/src/main/java/app/revanced/integrations/returnyoutubedislike/ReturnYouTubeDislike.java
```diff
@@ -394,15 +399,14 @@ private static void setSegmentedAdjustmentValues() {
                 segmentedLeftSeparatorHorizontalScaleRatio = 0.7f;
                 break;
             case "OnePlus":
-                configManufacturer = "oneplus";
+                configManufacturer = "OnePlus";
```
oSumAtrIX, line 402: Good catch, is this confirmed? I am doubting myself that I forgot to case it correctly.

app/src/main/java/app/revanced/integrations/settingsmenu/ReturnYouTubeDislikeSettingsFragment.java
```diff
@@ -80,6 +89,19 @@ public void onCreate(Bundle savedInstanceState) {
         });
         preferenceScreen.addPreference(percentagePreference);
 
+        compactLayoutPreference = new SwitchPreference(context);
```
oSumAtrIX, line 92: Preference screens such as this originate from legacy code by VancedTeam. Unrelated to this PR, but they should be converted to ReVanced style of settings.

oSumAtrIX (approved):

app/src/main/java/app/revanced/integrations/returnyoutubedislike/ReturnYouTubeDislike.java
LisoUseInAIKyrios: It must check the actual layout string, because during layout the method is called multiple times on the same span.  And layout happens anytime the span first appears on screen (such as scrolling off screen, and then scrolling back up).

So it must check the actual span each time, and not just once per video load.

app/src/main/java/app/revanced/integrations/returnyoutubedislike/ReturnYouTubeDislike.java
LisoUseInAIKyrios: 46 pixels, vs 50 pixels.  Not much of a difference, no.  But it is more 'compact'.

app/src/main/java/app/revanced/integrations/returnyoutubedislike/ReturnYouTubeDislike.java
LisoUseInAIKyrios: configManufacturer is only for logging, so it doesn't matter if the casing is wrong.

app/src/main/java/app/revanced/integrations/settingsmenu/ReturnYouTubeDislikeSettingsFragment.java
LisoUseInAIKyrios: What should this be migrated to?

oSumAtrIX: The patch should use the `settings` patch as a dependency and integrate its settings into the app. The `settings` patch does not support "custom" preference components yet, which would be necessary in the case of the `sponsorblock` patch.

LisoUseInAIKyrios: It would need to handle the statistics info that is shown/not shown based on debug status, and the additional RYD switches that become non selectable when  RYD is not enabled (percentage, compact layout).

oSumAtrIX: Yes, the `settings` patch would need to take a custom transformer which patches can supply if necessary. Not relevant to this PR though, just something to consider.

oSumAtrIX (approved):

PR #290, opened 2023-01-16
fix: microg check not showing toast

reisxd:
Fixes the Vanced MicroG check not showing a toast to make the user install Vanced MicroG

**Note: It still does not show a toast but it opens the download page for Vanced MicroG before crashing with this error:**

```
AndroidRuntime: FATAL EXCEPTION: yt-critical Thread #0
AndroidRuntime: Process: app.revanced.android.youtube, PID: 7179
AndroidRuntime: java.lang.SecurityException: Failed to find provider com.mgoogle.android.gsf.gservices for user 0; expected to find a valid ContentProvider for this authority
```

### DIFF
--- a/app/src/main/java/app/revanced/integrations/patches/MicroGSupport.java
+++ b/app/src/main/java/app/revanced/integrations/patches/MicroGSupport.java
@@ -26,6 +26,7 @@ public class MicroGSupport {
             Toast.makeText(context, str("microg_not_installed_warning"), Toast.LENGTH_LONG).show();
 
             var intent = new Intent(Intent.ACTION_VIEW);
+            intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
             intent.setData(Uri.parse(VANCED_MICROG_DOWNLOAD_LINK));
             context.startActivity(intent);
         }


app/src/main/java/app/revanced/integrations/patches/MicroGSupport.java
```diff
@@ -26,6 +26,7 @@ public static void checkAvailability() {
             Toast.makeText(context, str("microg_not_installed_warning"), Toast.LENGTH_LONG).show();
 
             var intent = new Intent(Intent.ACTION_VIEW);
+            intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
```
oSumAtrIX, line 29: Please clarify your PR and this line.

reisxd: In the check, the app would exit with the error `Calling startActivity() from outside of an Activity  context requires the  
FLAG_ACTIVITY_NEW_TASK flag. Is this really what you want?` and setting the flag `FLAG_ACTIVITY_NEW_TASK` would make the Vanced MicroG release link open in a browser. 

Source: https://stackoverflow.com/questions/3918517/calling-startactivity-from-outside-of-an-activity-context

PR #293, opened 2023-01-17
feat(YouTube) show toast on ReVanced error

LisoUseInAIKyrios:
Integration changes for https://github.com/revanced/revanced-patches/pull/1511

Display a toast if a ReVanced error occurs.

This includes normal and common exceptions (RYD or SponsorBlock connection timeout), but also includes showing a toast for all other exceptions (real problems that should never occur).

This should help users figure out when something is wrong, and will give more information to report instead of posting a screen shot of an empty black screen with a comment of: "it don't work. I don't know."

### DIFF
--- a/app/src/main/java/app/revanced/integrations/utils/LogHelper.java
+++ b/app/src/main/java/app/revanced/integrations/utils/LogHelper.java
@@ -55,6 +55,8 @@ public class LogHelper {
 
     /**
      * Logs debug messages with the most outer class name of the code that is calling this method.
+     * Whenever possible, the log string should be constructed entirely inside {@link LogMessage#buildMessageString()}
+     * so the performance cost of building strings is paid only if {@link SettingsEnum#DEBUG} is enabled.
      */
     public static void printDebug(LogMessage message) {
         if (SettingsEnum.DEBUG.getBoolean()) {
--- a/app/src/main/java/app/revanced/integrations/utils/LogHelper.java
+++ b/app/src/main/java/app/revanced/integrations/utils/LogHelper.java
@@ -25,10 +25,12 @@ public class LogHelper {
          * @return For outer classes, this returns {@link Class#getSimpleName()}.
          * For inner, static, or anonymous classes, this returns the simple name of the enclosing class.<br>
          * <br>
-         * For example, each of these classes return 'SomethingView':<br>
-         * com.company.SomethingView<br>
-         * com.company.SomethingView$StaticClass<br>
-         * com.company.SomethingView$1<br>
+         * For example, each of these classes return 'SomethingView':
+         * <code>
+         * com.company.SomethingView
+         * com.company.SomethingView$StaticClass
+         * com.company.SomethingView$1
+         * </code>
          */
         private String findOuterClassSimpleName() {
             var selfClass = this.getClass();
--- a/app/src/main/java/app/revanced/integrations/utils/LogHelper.java
+++ b/app/src/main/java/app/revanced/integrations/utils/LogHelper.java
@@ -52,13 +52,6 @@ public class LogHelper {
     private static final String REVANCED_LOG_PREFIX = "revanced: ";
 
     /**
-     * Logs information messages using the outer class name of the code calling this method.
-     */
-    public static void printInfo(@NonNull LogMessage message) {
-        Log.i(REVANCED_LOG_PREFIX + message.findOuterClassSimpleName(), message.buildMessageString());
-    }
-
-    /**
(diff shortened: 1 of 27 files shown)


LisoUseInAIKyrios:
Maybe this should be enabled by default?  If all is working correctly, then exception logging will not occur.

Setting this enabled by default could greatly help the user determine that something is broken.

LisoUseInAIKyrios:
Maybe this should not be an enum, and instead it should always show a toast on an error?  Remove the enum entirely and always show a toast?

I cannot think of any realistic reason why anyone would not want to show a toast on an exception.

LisoUseInAIKyrios:
Maybe this should not be a setting, and instead its always on?

There is not a strong reason to ever turn this off.

LisoUseInAIKyrios:
And sure enough, this has already exposed an error that has gone unnoticed:

bug

```
revanced: ReVancedSettingActivity: Couldn't set Toolbar title
revanced: ReVancedSettingActivity: java.lang.NullPointerException: Attempt to invoke virtual method 'void android.widget.TextView.setText(java.lang.CharSequence)' on a null object reference
revanced: ReVancedSettingActivity: 	at app.revanced.integrations.settingsmenu.ReVancedSettingActivity.initializeSettings(ReVancedSettingActivity.java:47)

```

app/src/main/java/app/revanced/integrations/utils/LogHelper.java
```diff
@@ -13,11 +18,12 @@ public class LogHelper {
...
      */
     public interface LogMessage {
+        @NonNull
         String buildMessageString();
 
         /**
-         * @return For non inner classes, this returns {@link Class#getSimpleName()}.
-         * For inner classes (static and anonymous), this returns the enclosing class simple name.<br>
+         * @return For non nested classes, this returns {@link Class#getSimpleName()}.
+         * For classes of inner, static, or anonymous, this returns the simple name of the enclosing class.<br>
```
oSumAtrIX, line 26: For classes of inner, static, or anonymous _what_. This is missing here.

oSumAtrIX, line 26: Either use nested or inner, not both.

app/src/main/java/app/revanced/integrations/utils/LogHelper.java
```diff
@@ -64,58 +74,50 @@ public static void printDebug(LogMessage message) {
...
+        }
+        ReVancedUtils.runOnMainThread(() -> {
+            Context context = ReVancedUtils.getContext();
+            if (context != null) {
+                String toastMessageToDisplay = (userToastMessage != null)
+                        ? userToastMessage
+                        : outerClassSimpleName + ": " + messageString;
+                Toast.makeText(context, toastMessageToDisplay, Toast.LENGTH_LONG).show();
+            }
+        });
```
oSumAtrIX, line 120: This might be unwanted or at least should be toggleable. Imagine the RYD servers are down and you receive an annoying toast on each video.

oSumAtrIX:
> And sure enough, this has already exposed an error that has gone unnoticed:
> 
> bug
> 
> ```
> revanced: ReVancedSettingActivity: Couldn't set Toolbar title
> revanced: ReVancedSettingActivity: java.lang.NullPointerException: Attempt to invoke virtual method 'void android.widget.TextView.setText(java.lang.CharSequence)' on a null object reference
> revanced: ReVancedSettingActivity: 	at app.revanced.integrations.settingsmenu.ReVancedSettingActivity.initializeSettings(ReVancedSettingActivity.java:47)
> ```

This exception is already known and tracked with revanced/revanced-patches/issues/1384. Make sure exception toasts are toggleable otherwise it can get quite annoying :)

app/src/main/java/app/revanced/integrations/utils/LogHelper.java
LisoUseInAIKyrios: Connection errors are the only realistic reason anyone would ever turn the toasts off.

I restored the setting switch, if someone really wants the toasts gone.

oSumAtrIX: Any other exception which causes the toast to show up multiple times in short periods is reason to as well.

LisoUseInAIKyrios:
before this PR can be pulled in, bug https://github.com/revanced/revanced-integrations/issues/300 should be fixed.  Otherwise if someone only uses shorts, then every like/dislike will show a toast.

oSumAtrIX (approved):

PR #295, opened 2023-01-20
feat(Youtube): `open-links-externally` patch

johnconner122:


app/src/main/java/app/revanced/integrations/patches/OpenLinksExternallyPatch.java
```diff
@@ -0,0 +1,10 @@
+package app.revanced.integrations.patches;
+
+import app.revanced.integrations.settings.SettingsEnum;
+
+public class OpenLinksExternallyPatch {
+    public static String enableExternalBrowser(String original) {
+        if (SettingsEnum.ENABLE_EXTERNAL_BROWSER.getBoolean()) original = "";
+        return original;
+    }
```
oSumAtrIX, line 9: Please add a comment to this method. Explain why it is required to set `original` to an empty value.

oSumAtrIX (approved):

PR #297, opened 2023-01-20
feat: spoof-wifi-connection patch

Linus789:
Related patches PR: [revanced-patches#1527](https://github.com/revanced/revanced-patches/pull/1527)

### DIFF
--- a/app/src/main/java/app/revanced/all/connectivity/wifi/spoof/SpoofWifiPatch.java
+++ b/app/src/main/java/app/revanced/all/connectivity/wifi/spoof/SpoofWifiPatch.java
@@ -14,7 +14,7 @@ import androidx.annotation.RequiresApi;
 
 public class SpoofWifiPatch {
 
-    // Used in order to check what the (real or fake) active network is (take a look at `hasTransport`)
+    // Used to check what the (real or fake) active network is (take a look at `hasTransport`).
     private static ConnectivityManager CONNECTIVITY_MANAGER;
 
     // If Wifi is not enabled, these are types that would pretend to be Wifi for android.net.Network (lower index = higher priority).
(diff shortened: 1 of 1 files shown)


app/src/main/java/app/revanced/all/connectivity/fakewifi/FakeWifiPatch.java
```diff
@@ -0,0 +1,380 @@
+package app.revanced.all.connectivity.fakewifi;
```
oSumAtrIX, line 1: This class should be in the package `app.revanced.all.connectivity.wifi.fake`. Make sure to apply this in the related patch as well.

oSumAtrIX (commented):
Many code blocks contain a lot of duplicate code. This might be intentional to accommodate for readability and logical separation between the separate methods. If that is not the case, duplicate code should be consolidated.

Linus789:
Took a stab at removing some duplicate logic.

app/src/main/java/app/revanced/all/connectivity/wifi/spoof/SpoofWifiPatch.java
```diff
@@ -0,0 +1,407 @@
...
+import android.net.NetworkInfo;
+import android.net.NetworkRequest;
+import android.os.Build;
+import android.os.Handler;
+
+import androidx.annotation.RequiresApi;
+
+public class SpoofWifiPatch {
+
+    // Used in order to check what the (real or fake) active network is (take a look at `hasTransport`)
```
oSumAtrIX, line 17:
```suggestion
    // Used in order to check what the (real or fake) active network is (take a look at `hasTransport`).
```

oSumAtrIX (commented):
There is still a fair chunk of duplicate code left, though reducing it would take a toll on readability. For that reason, I'd say this is ready to get merged if you agree with me.

Linus789:
> There is still a fair chunk of duplicate code left, though reducing it would take a toll on readability. For that reason, I'd say this is ready to get merged if you agree with me.

I agree.

oSumAtrIX (dismissed):

PR #298, opened 2023-01-21
fix(YouTube) settings enum fails to save non-boolean values

LisoUseInAIKyrios:
Fix for settings not persistently saving across app sessions.

This might help the RYD backend, as the client was creating a new RYD user on the first vote of every app session.

### DIFF
--- a/app/src/main/java/app/revanced/integrations/settings/SettingsEnum.java
+++ b/app/src/main/java/app/revanced/integrations/settings/SettingsEnum.java
@@ -343,19 +343,19 @@ public enum SettingsEnum {
 
         switch (getReturnType()) {
             case FLOAT:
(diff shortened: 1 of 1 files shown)


oSumAtrIX (approved):
Good catch

SodaWithoutSparkles:
Waiting for a stable release to test in RV manager. By the looks of things it seems that this will do the trick.

SodaWithoutSparkles:
@LisoUseInAIKyrios I believe the `:` is missing for semantic release. Try changing the title to `fix(YouTube): settings enum fails to save non-boolean values`. Or add a new commit following semantic rules.

Reference: https://gist.github.com/joshbuchea/6f47e86d2510bce28f8e7f42ae84c716

PR #301, opened 2023-01-22
feat(youtube/microg-support): check if microg is running

reisxd:
Implemets a check to check if Vanced MicroG is running or not and if it's not running, it'll open the Don't Kill My App guide and show a toast.

### DIFF
--- a/app/src/main/java/app/revanced/integrations/patches/MicroGSupport.java
+++ b/app/src/main/java/app/revanced/integrations/patches/MicroGSupport.java
@@ -2,10 +2,13 @@ package app.revanced.integrations.patches;
 
 import static app.revanced.integrations.sponsorblock.StringRef.str;
 
+import android.content.Context;
 import android.content.Intent;
 import android.content.pm.PackageManager;
 import android.net.Uri;
 import android.widget.Toast;
+
+import java.util.Objects;
 
 import app.revanced.integrations.utils.LogHelper;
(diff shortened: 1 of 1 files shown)


app/src/main/java/app/revanced/integrations/patches/MicroGSupport.java
```diff
@@ -14,13 +14,28 @@ public class MicroGSupport {
...
             LogHelper.printDebug(() -> "MicroG is installed on the device");
+
+            if (microGClient == null) {
+                // This means that ReVanced cannot access MicroG, probably because
+                // it's not running in the background.
+                
+                Toast.makeText(context, str("microg_not_running_warning"), Toast.LENGTH_LONG).show();
+                var intent = new Intent(Intent.ACTION_VIEW);
+                intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+                intent.setData(Uri.parse(DONT_KILL_MY_APP_LINK));
```
oSumAtrIX, line 36: These shouldn't throw any exceptions; please only try catching exceptions from  `context.startActivity(intent);`.

reisxd: I don't think it'd work correctly. If MicroG isn't installed, it'd launch both the Vanced MicroG release page and Don't kill my app guide.

oSumAtrIX (approved):

PR #302, opened 2023-01-22
fix(YouTube) ReturnYouTubeDislike shorts voting fix

LisoUseInAIKyrios:
Integration changes for https://github.com/revanced/revanced-patches/pull/1532

### DIFF
--- a/app/src/main/java/app/revanced/integrations/shared/PlayerType.kt
+++ b/app/src/main/java/app/revanced/integrations/shared/PlayerType.kt
@@ -7,7 +7,7 @@ import app.revanced.integrations.utils.Event
  */
 @Suppress("unused")
 enum class PlayerType {
-    NONE, // this also includes when a Shorts is playing
+    NONE, // this also includes when shorts are playing
     HIDDEN,
     WATCH_WHILE_MINIMIZED,
     WATCH_WHILE_MAXIMIZED,
--- a/app/src/main/java/app/revanced/integrations/shared/PlayerType.kt
+++ b/app/src/main/java/app/revanced/integrations/shared/PlayerType.kt
@@ -7,7 +7,7 @@ import app.revanced.integrations.utils.Event
  */
 @Suppress("unused")
 enum class PlayerType {
-    NONE,
+    NONE, // -- this also includes when a Shorts is playing --
     HIDDEN,
     WATCH_WHILE_MINIMIZED,
     WATCH_WHILE_MAXIMIZED,
--- a/app/src/main/java/app/revanced/integrations/shared/PlayerType.kt
+++ b/app/src/main/java/app/revanced/integrations/shared/PlayerType.kt
@@ -7,7 +7,7 @@ import app.revanced.integrations.utils.Event
  */
 @Suppress("unused")
 enum class PlayerType {
-    NONE, // -- this also includes when a Shorts is playing --
+    NONE, // this also includes when a Shorts is playing
     HIDDEN,
     WATCH_WHILE_MINIMIZED,
     WATCH_WHILE_MAXIMIZED,
--- a/app/src/main/java/app/revanced/integrations/sponsorblock/PlayerController.java
+++ b/app/src/main/java/app/revanced/integrations/sponsorblock/PlayerController.java
@@ -58,6 +58,8 @@ public class PlayerController {
             return;
         }
 
+        // currently this runs every time a video is loaded (regardless if sponsorblock is turned on or off)
+        // FIXME: change this so if sponsorblock is disabled, then run this method exactly once and once only
         SponsorBlockSettings.update(null);
 
         if (!SettingsEnum.SB_ENABLED.getBoolean()) {
--- a/app/src/main/java/app/revanced/integrations/sponsorblock/PlayerController.java
+++ b/app/src/main/java/app/revanced/integrations/sponsorblock/PlayerController.java
@@ -20,6 +20,7 @@ import java.util.TimerTask;
 
 import app.revanced.integrations.patches.VideoInformation;
 import app.revanced.integrations.settings.SettingsEnum;
+import app.revanced.integrations.shared.PlayerType;
 import app.revanced.integrations.sponsorblock.objects.SponsorSegment;
 import app.revanced.integrations.sponsorblock.requests.SBRequester;
 import app.revanced.integrations.utils.LogHelper;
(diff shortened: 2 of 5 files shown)


app/src/main/java/app/revanced/integrations/sponsorblock/PlayerController.java
```diff
@@ -45,6 +45,10 @@ public class PlayerController {
 
     public static boolean shorts_playing = false;
 
+    public static void shortsOpened() {
+        shorts_playing = true;
+    }
+
```
oSumAtrIX, line 51: Previously the field `shorts_playing` was set directly by injecting the necessary instructions for that in the method when shorts open. Since this PR now makes use of a patch to detect if shorts open, this method was introduced instead. It was found that `shorts_playing` is never set back to false, which blocks `executeDownloadSegments`. I am not sure though, why it still works.

oSumAtrIX (approved):
Please test this and confirm if the refactoring changes work for you.

app/src/main/java/app/revanced/integrations/sponsorblock/PlayerController.java
LisoUseInAIKyrios: shorts_playing is set to false using borrowed code from the autocaptions patch:

```
        val startVideoInformerMethod = StartVideoInformerFingerprint.result!!.mutableMethod
        startVideoInformerMethod.addInstructions(
            0, """
            const/4 v0, 0x0
            sput-boolean v0, $INTEGRATIONS_PLAYER_CONTROLLER_CLASS_DESCRIPTOR->shorts_playing:Z
        """
        )
```
Although this could be removed.  Since VideoIdPatch does not support shorts, the call to PlayerController#setCurrentVideoId could detect when a non short is loaded.

LisoUseInAIKyrios:
Yes, it's working for both RYD and SponsorBlock.

app/src/main/java/app/revanced/integrations/sponsorblock/PlayerController.java
oSumAtrIX: > Although this could be removed. Since VideoIdPatch does not support shorts, the call to PlayerController#setCurrentVideoId could detect when a non short is loaded.

I do not understand this quite yet. Does this snippet not already suffice to determine if shorts are playing? If that is the case, `VideoInformationPatch` should be refactored to include a getter for when shorts are playing. This would allow other patches to depend on it to check if shorts are currently playing such as the patch for RYD and SponsorBlock, since SponsorBlock does not properly integrate with shorts. This refactoring change would get rid of the following code from the SponsorBlock patch:
 
```kt
        val startVideoInformerMethod = StartVideoInformerFingerprint.result!!.mutableMethod
        startVideoInformerMethod.addInstructions(
            0, """
            const/4 v0, 0x0
            sput-boolean v0, $INTEGRATIONS_PLAYER_CONTROLLER_CLASS_DESCRIPTOR->shorts_playing:Z
        """
        )

        val shortsPlayerConstructorMethod = ShortsPlayerConstructorFingerprint.result!!.mutableMethod

        shortsPlayerConstructorMethod.addInstructions(
            0, """
            const/4 v0, 0x1
            sput-boolean v0, $INTEGRATIONS_PLAYER_CONTROLLER_CLASS_DESCRIPTOR->shorts_playing:Z
        """
        )
```

Instead, SponsorBlock would depend on `VideoInformationPatch` and use its getter for `shortsPlaying`.

LisoUseInAIKyrios: Yes that sounds correct.

Additional, since VideoIdPatch does not currently support shorts, that means sponsorblock does not need to check if a short is playing since it will only be notified of regular video ids that are loaded.

So the two patch injections you quoted could be removed from sponsorblock without any problem.

oSumAtrIX: I am still confused why the `sponsorblock` patch injects code which sets `shorts_playing` to true or false in addition to the current shorts detection patch. Instead of the current approach, the VideoInformationPatch class should adopt that code from the `sponsorblock` patch and provide a getter for getting the current state of shorts playing. As explained, this would get rid of introducing the ShortsDetectionPatch class. Instead other patches would use VideoInformationPatch as a dependency so that their integration can make use of said getter. After all, it makes sense that shorts playing state is related to video information, so it belongs in that class.

LisoUseInAIKyrios: I agree

LisoUseInAIKyrios: I agree that shorts detection should be moved into VideoInformationPatch.

LisoUseInAIKyrios:
made some changes, and also fixed issue https://github.com/revanced/revanced-integrations/issues/305

Using the PlayerType setting, I was able to remove the entire need of the shorts detection patch. 

Definitely should keep the new patch, because it still works and maybe it'll be used in the future.

app/src/main/java/app/revanced/integrations/shared/PlayerType.kt
```diff
@@ -7,7 +7,7 @@ import app.revanced.integrations.utils.Event
  */
 @Suppress("unused")
 enum class PlayerType {
-    NONE,
+    NONE, // -- this also includes when a Shorts is playing --
```
oSumAtrIX, line 10:
```suggestion
    NONE, // this also includes when a Shorts is playing
```

app/src/main/java/app/revanced/integrations/shared/PlayerType.kt
```diff
@@ -7,7 +7,7 @@ import app.revanced.integrations.utils.Event
  */
 @Suppress("unused")
 enum class PlayerType {
-    NONE,
+    NONE, // this also includes when a Shorts is playing
```
oSumAtrIX, line 10:
```suggestion
    NONE, // this also includes when shorts are playing
```

oSumAtrIX (approved):

oSumAtrIX:
> Should still keep the new ShortsDetectionPatch, because it works and it might be integrated into the video information patch.

I believe it is okay to remove it because the `PlayerTypeHookPatch` already exists with which it is possible to determine if shorts are playing or not.

PR #307, opened 2023-01-28
fix(YouTube) error when switching on/off interaction setting

LisoUseInAIKyrios:
integration changes for https://github.com/revanced/revanced-patches/pull/1550


oSumAtrIX (approved):

oSumAtrIX:
Thanks for your continuous contribution. Since you use similar titles like `fix(YouTube) error when switching on/off interaction setting` you can check [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/), which our repositories follow. Conventionally, instead of  `fix(YouTube) error when switching on/off interaction setting`, the formatting should look like this: `fix(youtube): error when switching on/off interaction setting` with a `:` after the commit type and lowercase text (except for names in the message after `:`.

PR #311, opened 2023-02-07
fix(youtube): fix load settings

LisoUseInAIKyrios:
I changed the old Vanced code that attempted to load primitive settings as strings (which will never work, they are always stored as primitives).  It now loads the primitive type, and if the existing preference is the wrong type it overwrites with the default value of the correct type.


oSumAtrIX:
I intend to change how integrations work. All of this design is adapted from old Vanced, which wasn't intended for modularity. Ideally patches ship their own integrations classes (which can depend on other patches integrations or on arbitrary, patch bundle global integrations). This way only the integrations get merged which are required. The issue I currently have is that I am not sure yet, how to achieve this properly. The patches have to compile but ship their integrations as resources, while also allowing javac to compile and dex to transpile the classes separately. All of this requires proper discussion and ideas.

PR #312, opened 2023-02-09
feat(youtube/return-youtube-dislike): shorts support

LisoUseInAIKyrios:
- support for RYD shorts
- fixed UI not showing updated dislike values after disliking

required patch changes: https://github.com/revanced/revanced-patches/pull/1596


oSumAtrIX (approved):

app/src/main/java/app/revanced/integrations/patches/ReturnYouTubeDislikePatch.java
```diff
@@ -1,39 +1,63 @@
...
      *
      * @param vote -1 (dislike), 0 (none) or 1 (like)
      */
     public static void sendVote(int vote) {
-        for (ReturnYouTubeDislike.Vote v : ReturnYouTubeDislike.Vote.values()) {
-            if (v.value == vote) {
-                ReturnYouTubeDislike.sendVote(v);
-                return;
+        try {
+            if (!SettingsEnum.RYD_ENABLED.getBoolean()) return;
```
oSumAtrIX, line 50: This check is already being done here:

https://github.com/revanced/revanced-integrations/blob/076289eb1bf171843eeb45691bf032930dced85e/app/src/main/java/app/revanced/integrations/returnyoutubedislike/ReturnYouTubeDislike.java#L297

It might be desired to remove it from there or here.

app/src/main/java/app/revanced/integrations/shared/PlayerType.kt
```diff
@@ -51,7 +52,7 @@ enum class PlayerType {
     }
 
     /**
-     * Weather Shorts are being played.
+     * Useful to check for Shorts playback
```
oSumAtrIX, line 55:
```suggestion
     * Check if Shorts are being played.
```

oSumAtrIX:
```suggestion
     * Check if the current player type is ˋNONEˋ or ˋHIDDENˋ.
     * @return True, if nothing or Shorts are being played.

**Comment by oSumAtrIX** [comment]
That looks pretty good. Because the PRs need to get squashed before merge, unrelated commits to the PR such as the latest, should be done on a separate branch and PR so that it is possible to chose a suitable commit name.

**Comment by oSumAtrIX** [comment]
Does it really depend on this PR?

**Review by oSumAtrIX — CHANGES_REQUESTED** [review]
Seems like something went wrong and made some unwanted changes appear in this PR. Please rebase on top of/merge branch dev.

File: app/src/main/java/app/revanced/integrations/shared/PlayerType.kt [hunk 3]
```diff
@@ -51,7 +52,9 @@ enum class PlayerType {
     }
 
     /**
-     * Weather Shorts are being played.
+     * Check if the current player type is [NONE] or [HIDDEN]
+     *
+     * @return True, if nothing is playing or a Shorts or YouTube Story is playing.
```
oSumAtrIX, line 57:
```suggestion
     * @return True, if nothing, a Short, or a Story is playing.
```

app/src/main/java/app/revanced/integrations/returnyoutubedislike/ReturnYouTubeDislike.java
```diff
@@ -1,72 +1,88 @@
...
      * Must be less than 5 seconds, as per:
      * https://developer.android.com/topic/performance/vitals/anr
      */
-    private static final long MILLISECONDS_TO_BLOCK_UI_WHILE_WAITING_FOR_FETCH_VOTES_TO_COMPLETE = 4000;
+    private static final long MAX_MILLISECONDS_TO_BLOCK_UI_WHILE_WAITING_FOR_FETCH_VOTES_TO_COMPLETE = 4000;
+
+    /**
+     * Separator character to use for segmented like/dislike
+     */
+    private static final char middleSeparatorCharacter = '•';
```
oSumAtrIX, line 41: By convention finals are uppercase, is this casing intended here?

oSumAtrIX:
Those files are autogenerated/auto committed and should not be part of any PR.

oSumAtrIX:
No conflicts should arise

app/src/main/java/app/revanced/integrations/returnyoutubedislike/ReturnYouTubeDislike.java
```diff
@@ -163,67 +197,109 @@ public static void onComponentCreated(Object conversionContext, AtomicReference<
...
+        } catch (Exception ex) {
+            LogHelper.printException(() -> "Error while trying to update dislikes", ex);
+        }
+    }
 
-            if (updateDislike(textRef, isSegmentedButton, votingData)) {
-                LogHelper.printDebug(() -> "Updated dislike span to: " + textRef.get());
-            } else {
-                LogHelper.printDebug(() -> "Ignoring already updated dislike span: " + textRef.get());
+    public static Spanned onShortsComponentCreated(Spanned span) {
```
oSumAtrIX, line 209: Unlike onComponentCreated, this code does not need to check, if `conversionContextString(...)` contains the required span. How is this the case & can this be reflected to `conversionContextString(...)` as well? The parameter  `conversionContext` from the method `conversionContextString(...)`  actually is an object of which the `StringBuilder` field can be read from. This way the entire objects `toString()` override does not need to be called and instead the relevant string can be checked for the component name.

oSumAtrIX:
Thanks for your huge efforts!

PR #325, opened 2023-02-24
fix(youtube/return-youtube-dislike): improve segmented like/dislike layout

LisoUseInAIKyrios:
Changed the segmented like/dislike separators to use custom shape spans.

This gives better layout on all devices, and should improve the appearance on devices with unusual system fonts.

patch changes: https://github.com/revanced/revanced-patches/pull/1686


oSumAtrIX:
I'd consider this a fix, as this PR fixes the issue of layouts instead of changing the behavior of a feature, what do you think?

oSumAtrIX:
Please make sure to squash `feat` -> `dev`

oSumAtrIX:
I'll fix it later

PR #328, opened 2023-03-02
feat(youtube/settings): disable preference control if the feature is turned off.  Show a dialog explaining side effects of some patches

LisoUseInAIKyrios:
Added optional 'parent' field to SettingEnum.
Where the setting is dependent on the parent being enabled.

For example, `ENABLE_SWIPE_HAPTIC_FEEDBACK` cannot be changed in the Settings, unless either `ENABLE_SWIPE_BRIGHTNESS` or `ENABLE_SWIPE_VOLUME` is turned on.
disabled
enabled

Added optional user notice message, which shows a dialog if the  the user changes the setting from it's default.

[patch changes](https://github.com/revanced/revanced-patches/pull/1801)

### DIFF
--- a/app/src/main/java/app/revanced/integrations/settings/SettingsEnum.java
+++ b/app/src/main/java/app/revanced/integrations/settings/SettingsEnum.java
@@ -173,6 +173,9 @@ public enum SettingsEnum {
     public final SharedPrefCategory sharedPref;
     @NonNull
     public final ReturnType returnType;
+    /**
+     * If the app should be rebooted, if this setting is changed
+     */
     public final boolean rebootApp;
     /**
      * Set of boolean parent settings.
@@ -193,8 +196,8 @@ public enum SettingsEnum {
     @Nullable
     public final StringRef userDialogMessage;
 
-    // must be volatile, as some settings are read/write from different threads
-    // of note, the object value is persistently stored using SharedPreferences (which is thread safe)
+    // Must be volatile, as some settings are read/write from different threads.
+    // Of note, the object value is persistently stored using SharedPreferences (which is thread safe).
     @NonNull
     private volatile Object value;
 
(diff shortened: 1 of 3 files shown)


app/src/main/java/app/revanced/integrations/patches/HideReelsPatch.java
```diff
@@ -5,6 +5,7 @@
 import app.revanced.integrations.adremover.AdRemoverAPI;
 import app.revanced.integrations.settings.SettingsEnum;
 
+// edit: appears to be dead code
```
oSumAtrIX, line 8: Can be removed

oSumAtrIX:
I personally prefer the settings not to be disabled for the reason that you can change them while they are off, it is convenient when you decide to turn them on. With this PR, you first have to enable the setting, change the feature, and then disable it again, but I believe this is a better approach in terms of UX.

oSumAtrIX:
How would this PR be compatible with settings added via a patch? Take the debugging patch as an example, it comes with a switch to enable debugging but also provides the feature to log stack traces.

oSumAtrIX:
I am currently in the middle of exams, which is my time is limited on reviews until Saturday, then im free again for a considerable time. Please bear with me, thanks.

oSumAtrIX:
Why do swipe controls have two dependents? Sounds kinda odd to me.

app/src/main/java/app/revanced/integrations/settings/SettingsEnum.java
```diff
@@ -1,263 +1,377 @@
...
+    }
+    SettingsEnum(String path, ReturnType returnType, Object defaultValue, SharedPrefCategory prefName,
+                 boolean rebootApp, SettingsEnum[] parents) {
+        this(path, returnType, defaultValue, prefName, rebootApp, null, parents);
+    }
+    SettingsEnum(String path, ReturnType returnType, Object defaultValue, SharedPrefCategory prefName,
+                 SettingsEnum[] parents) {
+        this(path, returnType, defaultValue, prefName, false, null, parents);
+    }
+    SettingsEnum(String path, ReturnType returnType, Object defaultValue, SharedPrefCategory prefName,
```
oSumAtrIX, line 254: Is there a way to reduce this boilerplate?

oSumAtrIX (approved):

PR #331, opened 2023-03-14
fix(YouTube/GeneralAds): fix `general-ads` showing in feed

johnconner122:


oSumAtrIX (approved):

oSumAtrIX:
Would be interesting to have statistics on the block lists, maybe some old strings are not required anymore when YouTube updates the ads.

PR #337, opened 2023-03-15
feat(youtube/spoof-signature-verification): automatic signature spoofing

LisoUseInAIKyrios:
A temporary improvement for [side effects of signature spoofing](https://github.com/revanced/revanced-patches/issues/1752).

If a playback error occurs then spoofing is turned on.

**This PR needs verification using an account that is affected by the playback issue.**

[patch changes](https://github.com/revanced/revanced-patches/pull/1757)

### DIFF
--- a/app/src/main/java/app/revanced/integrations/patches/SpoofSignatureVerificationPatch.java
+++ b/app/src/main/java/app/revanced/integrations/patches/SpoofSignatureVerificationPatch.java
@@ -72,6 +72,7 @@ public class SpoofSignatureVerificationPatch {
                         ReVancedUtils.getContext(),
                         "Spoofing app signature to prevent playback issues", Toast.LENGTH_LONG
                 ).show();
+                // it would be great if the video could be forcefully reloaded, but currently there is no code to do this
             });
 
         } catch (Exception ex) {


oSumAtrIX:
> If a playback error occurs, then spoofing is turned on for the remainder of the app session.

It should persist between sessions and permanently enable, otherwise at every session, the first time the video wont load.

oSumAtrIX:
Looks like the fingerprint resolves to [this](https://androidx.tech/artifacts/media3/media3-datasource-cronet/1.0.0-alpha03-source/androidx/media3/datasource/cronet/CronetDataSource.java.html). I'll add some refactoring.

app/src/main/java/app/revanced/integrations/patches/SpoofSignatureVerificationPatch.java
```diff
@@ -1,40 +1,97 @@
...
+     * Injection point. Runs off the main thread.
+     *
+     * Used to check the status of http calls made to YouTube.
+     * Calls of interest have status code in the 4xx range (usually 403),
+     * which usually indicates a signature verification failure.
+     *
+     * @param statusCode           HTTP status code of the completed YouTube connection
+     * @param urlConnectionHeaders all connection headers of the completed connection
+     */
+    public static void connectionCompleted(int statusCode, Map<String, List<String>> urlConnectionHeaders) {
```
oSumAtrIX, line 71: What is the second parameter used for?

oSumAtrIX: Since this method hooks in a quite low-level method to make requests, if an unrelated request returns a matching exception, it could falsely detect playback issues.

oSumAtrIX: It is likely that all requests are highly abstracted, so some kind of hooking would be required to determine if the request belongs to the one, we are interested in. Maybe, instead of the URL, the request made to can be used to check, if we are currently hooking the correct request.

oSumAtrIX: `UrlResponseInfo.getUrl()` can be used for that.

app/src/main/java/app/revanced/integrations/patches/SpoofSignatureVerificationPatch.java
```diff
@@ -1,40 +1,102 @@
...
+
+            if (SettingsEnum.SIGNATURE_SPOOFING.getBoolean()) {
+                return; // already enabled
+            }
+
+            if (responseCode < 400 || responseCode >= 500){
+                return; // everything normal
+            }
+
+            // TODO: check for the current url using UrlResponseInfo.getUrl()
```
oSumAtrIX, line 79:
```suggestion
```

app/src/main/java/app/revanced/integrations/patches/SpoofSignatureVerificationPatch.java
```diff
@@ -1,40 +1,101 @@
...
+
+            if (SettingsEnum.SIGNATURE_SPOOFING.getBoolean()) {
+                return; // already enabled
+            }
+
+            if (responseCode < 400 || responseCode >= 500){
+                return; // everything normal
+            }
+
+
```
oSumAtrIX, line 79:
```suggestion
```

app/src/main/java/app/revanced/integrations/patches/SpoofSignatureVerificationPatch.java
oSumAtrIX: In that case, no check for the URL or domain is required and can be removed.

PR #344, opened 2023-04-02
fix(youtube): updated http user agent

LisoUseInAIKyrios:
Updated the user agent string


oSumAtrIX (approved):

oSumAtrIX:
@LisoUseInAIKyrios Thanks for this PR. A suggestion for the future:

This PR fixes an issue with the integrations for the patch `sponsorblock` for YouTube. For that reason, instead of `fix(youtube): updated http user agent` a better commit message would be `fix(youtube/sponsorblock): use correct user agent for requests`.

Note how we remain in the present tense (use), instead of past tense (update~~d~~) and it is clear what has been changed (corrected a wrong user agent for requests).

PR #346, opened 2023-04-03
fix(youtube/spoof-signature-verification): additional fixes for subtitle locations

LisoUseInAIKyrios:
I saw there is a third set of incorrect subtitle parameters.

The subtitle fix could always override non-shorts with a single default value (34, 50, 95), but that would override any custom subtitle positions a video may have.

For now, this PR seems to fix all situations.

### DIFF
--- a/app/src/main/java/app/revanced/integrations/patches/SpoofSignatureVerificationPatch.java
+++ b/app/src/main/java/app/revanced/integrations/patches/SpoofSignatureVerificationPatch.java
@@ -125,10 +125,10 @@ public class SpoofSignatureVerificationPatch {
      * Known incorrect default Shorts subtitle parameters, and the corresponding correct (non-Shorts) values.
      */
     private enum SubtitleWindowReplacementSettings {
-        DEFAULT_SHORTS_PARAMETERS_1(9, 20, 0, true, false,
-                34, 50, 90),
-        DEFAULT_SHORTS_PARAMETERS_2(10, 50, 0, true, false,
+        DEFAULT_SHORTS_PARAMETERS_1(10, 50, 0, true, false,
                 34, 50, 95),
+        DEFAULT_SHORTS_PARAMETERS_2(9, 20, 0, true, false,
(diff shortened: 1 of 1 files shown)


app/src/main/java/app/revanced/integrations/patches/SpoofSignatureVerificationPatch.java
```diff
@@ -76,62 +76,91 @@ public static void onResponse(int responseCode) {
...
+     * Known incorrect default Shorts subtitle parameters, and the corresponding (non-Shorts) correct values.
+     */
+    private static final SubtitleWindowReplacementSettings[] knownReplacements = {
+            new SubtitleWindowReplacementSettings(9, 20, 0, true, false,
+                    34, 50, 90),
+            new SubtitleWindowReplacementSettings(10, 50, 0, true, false,
+                    34, 50, 95),
+            new SubtitleWindowReplacementSettings(9, 20, 0, true, true,
+                    33, 20, 100),
+    };
```
oSumAtrIX, line 92: At this point probably an enum is a better choice.

oSumAtrIX (approved):

app/src/main/java/app/revanced/integrations/patches/SpoofSignatureVerificationPatch.java
oSumAtrIX: Can a name be inferred from the values? It is known what those values do, maybe a name can be inferred from that for each combination of values.

oSumAtrIX: Some name along these lines sounds fine to me.

SodaWithoutSparkles:
> [...] but that would override any custom subtitle positions a video may have.

Some videos might need this to differentiate between multiple speakers speaking at the same time, or not to block some important information that just happens to be at the bottom. Overriding any custom value might not be a good idea.

PR #355, opened 2023-04-11
fix(youtube/general-ads): remove broken ad filter

johnconner122:
It removes the `buy premium` banner but semi-transparent overlay is still there.
image

### DIFF
--- a/app/src/main/java/app/revanced/integrations/patches/GeneralAdsPatch.java
+++ b/app/src/main/java/app/revanced/integrations/patches/GeneralAdsPatch.java
@@ -64,7 +64,6 @@ public final class GeneralAdsPatch extends Filter {
                 "carousel_footered_layout",
                 "text_image_button_layout",
                 "primetime_promo",
-                "feature_grid_interstitial",
                 "product_details",
                 "brand_video_shelf"
         );


oSumAtrIX:
I am unable to reproduce this.

johnconner122:
> I am unable to reproduce this.

Force-stop app, clear cache, and use a vpn with US location. You should see, or partially blocked banners like this:




Sometimes youtube asks to try premium.

oSumAtrIX (commented):
I've not yet experienced this issue. Any idea how to reproduce? Otherwise I'd leave this until anyone reports an issue

johnconner122:
[related issue](https://www.reddit.com/r/revancedapp/comments/12w959l/revanced_darkens_on_open_cant_press_anything/)

oSumAtrIX:
This PR would restore the ad again; have you checked if the pathBuilder string contains anything meaningful to block? That way, maybe, in addition to the ad, the parent component can be added.

johnconner122:
> have you checked if the pathBuilder string contains anything meaningful to block? 

There was another string `carousel_interstitial`, but problem still remained!

0xrxL:
> > have you checked if the pathBuilder string contains anything meaningful to block?
> 
> There was another string `carousel_interstitial`, but problem still remained!

Sometimes could be something without a precise component name, but just with a series of property names.

This is in general immediately placed before the component you blocked, and works as a kind of container.

oSumAtrIX (approved):

PR #362, opened 2023-04-21
fix(youtube/sponsorblock): Include milliseconds in the voting menu

sebm253:
Milliseconds are essential for SponsorBlock segments.


oSumAtrIX (commented):
Have you checked if the step size is in the milliseconds?

caneleex:
I'm not quite sure what you mean

oSumAtrIX:
You just changed the formatting, but does the UI allow to select timestamps in the milliseconds range?

caneleex:
yes, both on (re)vanced and in the desktop extension - otherwise skips would be terrible

PR #363, opened 2023-04-21
fix:(spoof-wifi-connection): fixed crash when wifi is disabled

LisoUseInAIKyrios:
Turned off proguard optimizations and added `-keepattributes` flag which keeps required attributes such as `InnerClasses`.

This will fix https://github.com/revanced/revanced-patches/issues/1935,
and should also fix https://github.com/revanced/revanced-patches/issues/1583

ProGuard will continue to remove unused code in the libraries used by ReVanced.

### DIFF
--- a/app/proguard-rules.pro
+++ b/app/proguard-rules.pro
@@ -20,6 +20,8 @@
 # hide the original source file name.
 #-renamesourcefileattribute SourceFile
 -dontobfuscate
+-dontoptimize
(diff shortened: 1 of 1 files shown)


Linus789:
I just checked and can confirm that this fixes the mentioned spoof wifi issue. So thanks a lot @LisoUseInAIKyrios for fixing this long overdue issue! :)

PR #366, opened 2023-04-22
fix(youtube/hide-video-action-buttons): fix hide action buttons not working for some users

LisoUseInAIKyrios:
integration changes for https://github.com/revanced/revanced-patches/pull/1959

### DIFF
--- a/app/src/main/java/app/revanced/integrations/settings/SettingsEnum.java
+++ b/app/src/main/java/app/revanced/integrations/settings/SettingsEnum.java
@@ -74,7 +74,7 @@ public enum SettingsEnum {
     HIDE_LIKE_DISLIKE_BUTTON("revanced_hide_like_dislike_button", BOOLEAN, FALSE),
     HIDE_DOWNLOAD_BUTTON("revanced_hide_download_button", BOOLEAN, FALSE),
     HIDE_PLAYLIST_BUTTON("revanced_hide_playlist_button", BOOLEAN, FALSE),
-    HIDE_CLIP_BUTTON("revanced_hide_clip_button", BOOLEAN, TRUE),
+    HIDE_CLIP_BUTTON("revanced_hide_clip_button", BOOLEAN, FALSE, "revanced_hide_clip_button_user_dialog_message"),
     HIDE_ACTION_BUTTONS("revanced_hide_action_buttons", BOOLEAN, FALSE),
 
     // Layout settings


app/src/main/java/app/revanced/integrations/patches/ButtonsPatch.java
```diff
@@ -11,7 +11,7 @@ public ButtonsPatch() {
                 new BlockRule(SettingsEnum.HIDE_LIKE_DISLIKE_BUTTON, "|like_button", "dislike_button"),
                 new BlockRule(SettingsEnum.HIDE_DOWNLOAD_BUTTON, "download_button"),
                 new BlockRule(SettingsEnum.HIDE_PLAYLIST_BUTTON, "save_to_playlist_button"),
-                new BlockRule(SettingsEnum.HIDE_ACTION_BUTTONS, "ContainerType|video_action_button")
+                new BlockRule(SettingsEnum.HIDE_ACTION_BUTTONS, "ContainerType|video_action_button", "|CellType|CollectionType|CellType|ContainerType|button.eml|", "|clip_button.eml|")
```
oSumAtrIX, line 14: Wouldn't it be better to separate clip_button into its own?

oSumAtrIX (approved):

app/src/main/java/app/revanced/integrations/patches/ButtonsPatch.java
oSumAtrIX: The spoofing patch is a separate patch; therefore, if excluded, the disclaimer wouldn't mean anything. 

Instead a disclaimer for the following can be used:

> but hide clip would only work for users who get the updated layout

oSumAtrIX: How about a "Hide clip button". Why not just add the filter that applies to those not affected?

oSumAtrIX: The filter only includes the "|button.eml|" string, but you said it wouldn't work for some because of A/B. So what about adding the string, that works for them?

oSumAtrIX: Why would it only work for some users if we include the filter string for bucket A and bucket B?

oSumAtrIX: I am referring to the new layout. I just used the old layout and RYD as an analogy for supporting both buckets of A/B. The current filter only works for bucket A because only the filter string exists for A. In addition to that the string for bucket B could be added, that way the clip button could reliably hidden.

oSumAtrIX: Incognito might be in the same bucket as your user. Can also be client/device/location specific.

oSumAtrIX: Under what filter is the clip button being hidden for the "old" litho layout.

oSumAtrIX: I see. In that case, a disclaimer should be added. The disclaimer should be added as a dialog, just like the spoof toggle.

oSumAtrIX (approved):

PR #372, opened 2023-04-27
fix(youtube): no longer need to restart the app after changing `copy-video-url` or `downloads` patch

LisoUseInAIKyrios:
No longer need to restart the app after enabling/disabling `copy-video-url` or `downloads`

Can now turn the feature on or off while a video is playing, and it all just works.


oSumAtrIX (approved):
Moved the animation fields out of the instance constructor. This might be reverted in case the context is null at the point this class is loaded. Instead a lazy approach could also be used.

PR #384, opened 2023-05-02
fix(youtube/return-youtube-dislike): fix potential error toast when using old UI layout

LisoUseInAIKyrios:
Apparently the dislike text view can be null?  Not sure when this would happen, but check for it anyways.


oSumAtrIX:
Likely, the hook can be moved somewhere where a null check exists, which would've spared the additional check-in integrations, but isn't a problem now.

PR #388, opened 2023-05-04
feat(youtube): import / export of revanced settings

LisoUseInAIKyrios:
integration changes for https://github.com/revanced/revanced-patches/pull/2077

### DIFF
--- a/app/src/main/java/app/revanced/integrations/settings/SettingsEnum.java
+++ b/app/src/main/java/app/revanced/integrations/settings/SettingsEnum.java
@@ -731,6 +731,11 @@ public enum SettingsEnum {
     // Begin import / export
 
     /**
+     * If a setting path has this prefix, then remove it before importing/exporting.
+     */
+    private static final String OPTIONAL_REVANCED_SETTINGS_PREFIX = "revanced_";
+
+    /**
      * The path, minus any 'revanced' prefix to keep json concise.
      */
     private String getImportExportKey() {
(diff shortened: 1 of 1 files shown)


app/src/main/java/app/revanced/integrations/patches/GeneralAdsPatch.java
```diff
@@ -16,45 +16,45 @@ public final class GeneralAdsPatch extends Filter {
     };
 
     private final BlockRule custom = new CustomBlockRule(
-            SettingsEnum.ADREMOVER_CUSTOM_ENABLED,
-            SettingsEnum.ADREMOVER_CUSTOM_REMOVAL
+            SettingsEnum.ADREMOVER_CUSTOM,
+            SettingsEnum.ADREMOVER_CUSTOM_FILTERS
```
oSumAtrIX, line 20: You can remove the `ADREMOVER` prefix if we are at it. It was a remnant from Vanceds decompiled code.

oSumAtrIX: Currently, we use `HIDE`, `DISABLE` or `ENABLE` prefix. Example:

- `HIDE_GENERAL_ADS`
- `HIDE_PAID_CONTENT`
- `ENABLED_CUSTOM_FILTERS`

oSumAtrIX: I don't think sorting is necessarily something we need or ever considered a requirement for this. If we want to group together settings, we should add an explicit way to do so instead of abusing the naming convention. An initial approach could be using categories.

PR #393, opened 2023-05-09
feat(youtube/video-speed): change custom video speeds inside app settings

LisoUseInAIKyrios:
integration changes for https://github.com/revanced/revanced-patches/pull/2114


app/src/main/java/app/revanced/integrations/patches/playback/speed/CustomVideoSpeedPatch.java
```diff
@@ -1,12 +1,80 @@
...
+                videoSpeeds[i] = speed;
+            }
+        } catch (Exception ex) {
+            LogHelper.printInfo(() -> "parse error", ex);
+            resetCustomSpeeds("Invalid custom video speeds. Using default values.");
+            loadSpeeds();
+        }
+    }
+
+    private static boolean arrayContains(float[] array, float value) {
```
oSumAtrIX, line 74: Iirc this method is a duplicate of another somewhere else and should instead be refactored to the existing utils class.

PR #394, opened 2023-05-09
fix(youtube/spoof-app-version): restore watch history preview

LisoUseInAIKyrios:
When spoofing to an old version, the`hide-breaking-news-shelf` patch incorrectly hides the watch history thumbnail previews.

As far as I know, the breaking news components are not present when spoofing to an old version of the app. So this should be a safe fix.


oSumAtrIX (approved):

oSumAtrIX:
Keep in mind, both patches are sperate. This fix breaks the modular nature of ReVanced.

PR #407, opened 2023-05-20
feat(reddit): add `sanitize-sharing-links` patch

kazimmt:
Related to https://github.com/revanced/revanced-patches/pull/2192

### DIFF
--- a/app/src/main/java/app/revanced/reddit/patches/SanitizeUrlQueryPatch.java
+++ b/app/src/main/java/app/revanced/reddit/patches/SanitizeUrlQueryPatch.java
@@ -1,0 +1,9 @@
+package app.revanced.reddit.patches;
+
+public class UrlSanitizer {
+    public static String removeTrackingParameters(String url) {
+        Pattern pattern = Pattern.compile(".?utm_source=.+")
+        Matcher matcher = pattern.matcher(url)
+        return matcher.replaceAll("")
+    }
+}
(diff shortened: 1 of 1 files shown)


app/src/main/java/app/revanced/reddit/patches/SanitizeUrlQueryPatch.java
```diff
@@ -0,0 +1,9 @@
+package app.revanced.reddit.patches;
+
+public class UrlSanitizer {
+    public static String removeTrackingParameters(String url) {
+        Pattern pattern = Pattern.compile(".?utm_source=.+")
+        Matcher matcher = pattern.matcher(url)
+        return matcher.replaceAll("")
+    }
```
oSumAtrIX, line 8: Please test the changes, also instead of a regex, parse the URL and remove the query from the URL:
```suggestion
public final class SanitizeUrlQueryPatch {
	private static final Pattern SANITIZE_PATTERN = Pattern.compile(".?utm_source=.+");
    public static String removeTrackingParameters(final String url) {
        return SANITIZE_PATTERN.matcher(url).replaceAll("");
    }
```

oSumAtrIX:
I made some refactoring changes; please test if this works as expected.

kazimmt:
> I made some refactoring changes; please test if this works as expected.

I Test it
& It's Working as expected

PR #409, opened 2023-05-25
feat(youtube/hide-seekbar): more fine grained hiding of seekbar

LisoUseInAIKyrios:
integration changes for https://github.com/revanced/revanced-patches/pull/2252

### DIFF
--- a/app/src/main/java/app/revanced/integrations/patches/theme/SeekbarColorPatch.java
+++ b/app/src/main/java/app/revanced/integrations/patches/theme/SeekbarColorPatch.java
@@ -56,7 +56,7 @@ public final class SeekbarColorPatch {
      * Injection point.
      *
      * Overrides all Litho components that use the YouTube seekbar color.
-     * Used only for the seekbar shown in view thumbnails.
+     * Used only for the video thumbnails seekbar.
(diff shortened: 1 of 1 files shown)


app/src/main/java/app/revanced/integrations/patches/theme/SeekbarColorPatch.java
```diff
@@ -51,12 +51,31 @@ public static int getCustomSeekbarColor() {
...
+     */
+    public static int getLithoColorOverride(int colorValue) {
+        if (colorValue == ORIGINAL_SEEKBAR_COLOR) {
+            if (SettingsEnum.HIDE_SEEKBAR_THUMBNAIL.getBoolean()) {
+                return 0x00000000;
+            }
+            return getSeekbarColorValue(ORIGINAL_SEEKBAR_COLOR);
+        }
+        return colorValue;
+    }
```
oSumAtrIX, line 71: I believe the methods were separated so that the niche option to color only while the player is focused exists. It may be refactored back to it's original implementation according to https://github.com/revanced/revanced-patches/pull/2252#issuecomment-1562808600.

PR #415, opened 2023-05-26
feat(youtube/hide-shorts-components): hide shorts info panel

LisoUseInAIKyrios:
integration changes for https://github.com/revanced/revanced-patches/pull/2278

### DIFF
--- a/app/src/main/java/app/revanced/integrations/patches/components/ShortsFilter.java
+++ b/app/src/main/java/app/revanced/integrations/patches/components/ShortsFilter.java
@@ -10,7 +10,6 @@ import android.view.View;
 import com.google.android.libraries.youtube.rendering.ui.pivotbar.PivotBar;
 
 import app.revanced.integrations.settings.SettingsEnum;
-import app.revanced.integrations.utils.LogHelper;
 
 public final class ShortsFilter extends Filter {
     public static PivotBar pivotBar;


app/src/main/java/app/revanced/integrations/patches/components/AdsFilter.java
```diff
@@ -60,7 +60,8 @@ public AdsFilter() {
 
         final var medicalPanel = new StringFilterGroup(
                 SettingsEnum.HIDE_MEDICAL_PANELS,
-                "medical_panel"
+                "medical_panel",
+                "shorts_info_panel_overview"
```
oSumAtrIX, line 64: I don't think this component is soley used for medical information but information about the short in general, thus should be moved.

oSumAtrIX: I have seen the info panel link to Wikipedia once on normal videos.

app/src/main/java/app/revanced/integrations/patches/components/AdsFilter.java
```diff
@@ -76,7 +76,8 @@ public AdsFilter() {
         final var infoPanel = new StringFilterGroup(
                 SettingsEnum.HIDE_HIDE_INFO_PANELS,
                 "publisher_transparency_panel",
-                "single_item_information_panel"
+                "single_item_information_panel",
+                "shorts_info_panel_overview"
```
oSumAtrIX, line 80: How about adding this here, as well as in `ShortsComponetsFilter`? The reason for this is that this component is related to shorts. The question would be, if the user disables the switch to hide the component in the `shorts-componets` patch, it will not make any change unless they change it in the `hide-ads` patch toggle, so the toggles have to be able to turn each other off. Maybe a separate toggle can be added and used with the same name in both patches so that toggling one toggle toggles the other too. Before implementing, please let me know what you think about this, so that we can discuss how we want to implement it.

app/src/main/java/app/revanced/integrations/patches/components/AdsFilter.java
oSumAtrIX: > Or maybe this should be part of the existing HIDE_HIDE_INFO_PANELS? I think that might be better.

Imagine the following scenario:

The user watches Shorts and notices the info panel, so they check out ReVanced patches and notice a patch that can hide components of Shorts; now, the user includes the patch, but hiding the info panel (a shorts component) is not present. For that reason, both patches should somehow be able to toggle this component on or off to satisfy the current patches' semantics as best as possible.

app/src/main/java/app/revanced/integrations/patches/components/AdsFilter.java
oSumAtrIX: > but would be a bunch of work now

It is as simple as adding a separate preference screen with two preferences. One that hides the info panel, another that hides the shorts info panel. The same preference key for the shorts info panel switch can then be used for a switch in the shorts components preference screen; toggling either preference will turn both toggles on or off and successfully satisfy the problem semantically. 

On an unrelated note, currently, litho patches are hard to separate. If you think about it, the info panel is not part of hiding ads, yet it is added through that patch. I have not come up with a proper solution for this yet.

app/src/main/java/app/revanced/integrations/patches/components/ShortsFilter.java
```diff
@@ -55,6 +61,9 @@ boolean isFiltered(final String path, final String identifier, final byte[] prot
         if (reelChannelBar.check(path).isFiltered())
             if (this.pathFilterGroups.contains(path)) return true;
 
+        // Shorts info panel path appears outside of reelChannelBar path.
+        if (infoPanel.isEnabled() && infoPanel.check(path).isFiltered()) return true;
```
oSumAtrIX, line 65: IIrc `check` already checks for `isEnabled`.

app/src/main/java/app/revanced/integrations/settings/SettingsEnum.java
```diff
@@ -138,6 +138,7 @@ public enum SettingsEnum {
     HIDE_SHORTS_COMMENTS_BUTTON("revanced_hide_shorts_comments_button", BOOLEAN, FALSE),
     HIDE_SHORTS_REMIX_BUTTON("revanced_hide_shorts_remix_button", BOOLEAN, FALSE),
     HIDE_SHORTS_SHARE_BUTTON("revanced_hide_shorts_share_button", BOOLEAN, FALSE),
+    HIDE_SHORTS_INFO_PANEL("revanced_hide_shorts_info_panel", BOOLEAN, FALSE),
```
oSumAtrIX, line 141: It may be true by default. Currently, everything is false for the sole reason, including any filter patch, it also includes all filters and causes unintended patches. A way to solve this as elegantly as possible would be to make the `LithoFilter` patch inject separate hooks for each type of filter patch. This would solve the current (big) problem of the litho patches being inseparable.

oSumAtrIX (approved):

app/src/main/java/app/revanced/integrations/patches/components/ShortsFilter.java
oSumAtrIX: I believe there was a motivation behind this. I think I had to use `check` somewhere despite the filter being enabled. Either way, looks good as is.

app/src/main/java/app/revanced/integrations/settings/SettingsEnum.java
oSumAtrIX: Yes, every filter can be enabled or disabled by default. The problem I mention here is regarding the following situation:

You include a <filter> patch such as "hide-ads". You expect it to hide ads. For some reason, now shorts components are hidden. The reason for this is, that including the "hide-ads" patch, causes LithoFilterPatch to inject a hook and thus enable all filters. An elegant solution is mentioned in my previous comment.

app/src/main/java/app/revanced/integrations/settings/SettingsEnum.java
```diff
@@ -138,7 +138,7 @@ public enum SettingsEnum {
     HIDE_SHORTS_COMMENTS_BUTTON("revanced_hide_shorts_comments_button", BOOLEAN, FALSE),
     HIDE_SHORTS_REMIX_BUTTON("revanced_hide_shorts_remix_button", BOOLEAN, FALSE),
     HIDE_SHORTS_SHARE_BUTTON("revanced_hide_shorts_share_button", BOOLEAN, FALSE),
-    HIDE_SHORTS_INFO_PANEL("revanced_hide_shorts_info_panel", BOOLEAN, FALSE),
+    HIDE_SHORTS_INFO_PANEL("revanced_hide_shorts_info_panel", BOOLEAN, TRUE),
```
oSumAtrIX, line 141: I personally think, most people would also want other shorts components to be hidden by default such as the remix or join component as they are almost not used in most cases. The main reason everything is set to false was motivated by the previously mentioned issue. Setting this to true causes that issue. If we want to ignore the issue for now, I think we can hide remix, join & thanks by default to (in another PR).

PR #427, opened 2023-06-26
fix: make indexOf return earlier

jimman2003:
No need to compute the failure function if not data is present!


oSumAtrIX (changes requested):

app/src/main/java/app/revanced/integrations/patches/components/LithoFilterPatch.java
```diff
@@ -83,9 +83,10 @@ public CustomFilterGroup(final SettingsEnum setting, final SettingsEnum filter)
 class ByteArrayFilterGroup extends FilterGroup<byte[]> {
     // Modified implementation from https://stackoverflow.com/a/1507813
     private int indexOf(final byte[] data, final byte[] pattern) {
+        if (data.length == 0)
+            return -1;
```
oSumAtrIX, line 87:
```suggestion
            return -1;

```

oSumAtrIX (approved):

PR #435, opened 2023-07-09
fix(youtube/spoof-signature-verification): show video time and chapters while using seekbar

LisoUseInAIKyrios:
integration changes for https://github.com/revanced/revanced-patches/pull/2607

### DIFF
--- a/app/src/main/java/app/revanced/integrations/patches/SpoofSignatureVerificationPatch.java
+++ b/app/src/main/java/app/revanced/integrations/patches/SpoofSignatureVerificationPatch.java
@@ -84,15 +84,11 @@ public class SpoofSignatureVerificationPatch {
      * @param view seekbar thumbnail view.  Includes both shorts and regular videos.
      */
     public static void seekbarImageViewCreated(ImageView view) {
-        try {
-            if (SettingsEnum.SPOOF_SIGNATURE_VERIFICATION.getBoolean()) {
-                view.setVisibility(View.GONE);
-                // Also hide the white border around the thumbnail (otherwise a 1 pixel wide bordered frame is visible).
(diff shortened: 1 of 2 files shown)


app/src/main/java/app/revanced/integrations/patches/SpoofSignatureVerificationPatch.java
```diff
@@ -67,4 +71,29 @@ public static String overrideProtobufParameter(String originalValue) {
...
+        try {
+            if (SettingsEnum.SPOOF_SIGNATURE_VERIFICATION.getBoolean()) {
+                view.setVisibility(View.GONE);
+                // Also hide the white border around the thumbnail (otherwise a 1 pixel wide bordered frame is visible).
+                ViewGroup parentLayout = (ViewGroup) view.getParent();
+                parentLayout.setPadding(0, 0, 0, 0);
+            }
+        } catch (Exception ex) {
+            LogHelper.printException(() -> "seekbarImageViewCreated failure", ex);
+        }
```
oSumAtrIX, line 96: Is the exception handler necessary at any time?

oSumAtrIX: Check if the view is null. If it is, return.

oSumAtrIX: My question is if there is a case where the view is null at any time. If it is, then the YouTube code checks for it somewhere. Moving the patch there would eliminate any checks. Otherwise, you can also show a toast when checking for null before you return. I can not think of any other unexpected situation. Do you have one?

oSumAtrIX (approved):

PR #416, opened 2023-05-28
feat(youtube): add `player-flyout-menu-panel` patch

johnconner122:
**TODO:**

- [ ]  Add `comments` filter to exceptions. These 3 comments' filters are hidden.

 
image

- [ ]  Test the feature for additional `unintended consequences`.

PR #447, opened 2023-07-24
perf(YouTube): Filter litho components using prefix tree

LisoUseInAIKyrios:
Since all litho filters are all searching against the same target path and protobuffer,  a prefix tree (trie) can be used to speed this up quite a bit.

This PR adds a prefix tree search class (`TrieSearch`), which works for both strings and byte arrays.

This default ReVanced patch installation uses 79+ filter strings, and a trie effectively searches for all of them in 1 pass.  This implementation does not prune away any non-enabled filters, so the runtime is basically a constant time regardless of how many filters are added or enabled.

Performance using default patches, default settings.
Average method runtime of each call to `LithoFilterPatch#filter` to run all enabled filters:
Before this PR, average method call runtime of 235,139 ns
With this PR, average method call runtime of 55,437 ns

Which gives a 425% speed up. Of note, that if more 'hide' features are enabled beyond the default settings then the speedup would be even greater (as the old linear searching would be adding more work, while the trie would continue to take the same time).   During development a 10x speedup of the worse case (slowest use cases) was seen.

The litho filter has one fundamental change, instead of each filter always being called with "Check if you match this path/protobuffer, then confirm if you want to hide the litho element".   With this PR the litho patch searches the trie once for matches, and then any matched filters are then confirmed with "I found a match of one of your filters, should I hide the litho component?"

Of note, the existing filter group classes still work the same as before, where they do a linear search for each pattern.  These existing filter groups can still be used to fine tune the filtering, just as ShortsFilter and a few other filters do.

### DIFF
--- a/app/src/main/java/app/revanced/integrations/patches/components/LithoFilterPatch.java
+++ b/app/src/main/java/app/revanced/integrations/patches/components/LithoFilterPatch.java
@@ -277,7 +277,7 @@ abstract class Filter {
      * If possible, avoid adding values to this list and instead use a path or identifier filter
      * for the item you are looking for. Then inside
      * {@link #isFiltered(String, String, byte[], FilterGroupList, FilterGroup, int)},
-     * the buffer can then be searched using using a
+     * the buffer can then be searched using using a different
      * {@link ByteArrayFilterGroupList} or a {@link ByteArrayFilterGroup}.
      * This way, the expensive buffer searching only occurs if the cheap and fast path/identifier is already found.
      */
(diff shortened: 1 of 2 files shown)


oSumAtrIX:
Cheers, in this case we can merge #416. It was put on hold due to performance issues

oSumAtrIX:
> Or should it only search from the current buffer index to the buffer limit?

I am unsure; giving the existing filters a try to search beginning at the current index would make sense. From what I have seen, the ByteBuffer object is streamed. And parsed by Protobuf. If we have to search the entire array, it may be sensible to add an option for filters to decide if they want to search from the beginning or the current index.

oSumAtrIX:
You could log the index for debugging purposes or check if the matching index is > currentIndex.

oSumAtrIX:
It may be possible to determine the start index to search for statistically. This would reduce the amortized runtime. Every time a match is found, save the index, calculate the average, and start doing a breadth-first search at the given index.

oSumAtrIX (approved):

PR #469, opened 2023-08-30
fix(YouTube/ExternalDownloads): Trim spaces from package

Autist69420:
If there were spaces in the package name anywhere, revanced would show a toast saying that the package could not be found, ideally it should find it no matter if there are spaces in the package name or not. This fixes that.

As discussed in discord, ideally package names should be trimmed on settings save but there is currently no abstraction for settings (as far as i can reinterpret from that message).

### DIFF
--- a/app/src/main/java/app/revanced/integrations/videoplayer/ExternalDownloadButton.java
+++ b/app/src/main/java/app/revanced/integrations/videoplayer/ExternalDownloadButton.java
@@ -49,7 +49,7 @@ public class ExternalDownloadButton extends BottomControlButton {
         LogHelper.printDebug(() -> "External download button clicked");
 
         final var context = view.getContext();
-        var downloaderPackageName = SettingsEnum.EXTERNAL_DOWNLOADER_PACKAGE_NAME.getString();
+        var downloaderPackageName = SettingsEnum.EXTERNAL_DOWNLOADER_PACKAGE_NAME.getString().trim();
 
         boolean packageEnabled = false;
         try {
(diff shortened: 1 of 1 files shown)


Ushie (approved):

app/src/main/java/app/revanced/integrations/videoplayer/ExternalDownloadButton.java
```diff
@@ -49,7 +49,7 @@ private static void onDownloadClick(View view) {
         LogHelper.printDebug(() -> "External download button clicked");
 
         final var context = view.getContext();
-        var downloaderPackageName = SettingsEnum.EXTERNAL_DOWNLOADER_PACKAGE_NAME.getString();
+        var downloaderPackageName = SettingsEnum.EXTERNAL_DOWNLOADER_PACKAGE_NAME.getString().trim();
```
oSumAtrIX, line 52:
```suggestion
        // Trim string to avoid any accidental whitespace.
        var downloaderPackageName = SettingsEnum.EXTERNAL_DOWNLOADER_PACKAGE_NAME.getString().trim();
```

oSumAtrIX (approved):

app/src/main/java/app/revanced/integrations/videoplayer/ExternalDownloadButton.java
Autist69420: Oh yeah I guess it would make sense to add a comment for that, I apologize it's 11 PM

PR #472, opened 2023-09-06
feat(TU Dortmund): Add `Show on lockscreen` patch

Traktores:
Integration changes for https://github.com/ReVanced/revanced-patches/pull/2947

### DIFF
--- a/app/src/main/java/app/revanced/tudortmund/lockscreen/ShowOnLockscreenPatch.java
+++ b/app/src/main/java/app/revanced/tudortmund/lockscreen/ShowOnLockscreenPatch.java
@@ -5,40 +5,38 @@ import android.hardware.display.DisplayManager;
 import android.os.Build;
 import android.view.Display;
 import android.view.Window;
-import android.view.WindowManager;
-
 import androidx.appcompat.app.AppCompatActivity;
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/app/revanced/tudortmund/lockscreen/ShowOnLockscreenPatch.java
```diff
@@ -0,0 +1,46 @@
...
+
+public class ShowOnLockscreenPatch {
+    /**
+     * @noinspection deprecation
+     */
+    public static Window getWindow(AppCompatActivity activity, float brightness) {
+        Window window = activity.getWindow();
+
+        if (brightness >= 0) {
+            // High brightness set, therefore show on lockscreen.
```
oSumAtrIX, line 21: Please explain what you mean by this.

Traktores: Whenever we open the student id or student ticket in the app, the brightness gets set to the maximum. That's why we check for when the brightness gets set and to what value. If it gets set to increase the brightness, we activate that the activity is shown on the lockscreen. And if the brightness gets reset, we disable it.

PR #479, opened 2023-09-23
refactor(Tumblr): Use a common filter patch

leumasme:
Required for the new Tumblr TimelineFilterPatch.  
See https://github.com/ReVanced/revanced-patches/pull/3018

### DIFF
--- a/app/src/main/java/app/revanced/tumblr/patches/TimelineFilterPatch.java
+++ b/app/src/main/java/app/revanced/tumblr/patches/TimelineFilterPatch.java
@@ -4,11 +4,10 @@ import com.tumblr.rumblr.model.TimelineObject;
 import com.tumblr.rumblr.model.Timelineable;
 
 import java.util.ArrayList;
-import java.util.Iterator;
 import java.util.List;
 
-public class TimelineFilterPatch {
-    static List<String> blockedObjectTypes = new ArrayList<>();
+public final class TimelineFilterPatch {
+    private static final List<String> blockedObjectTypes = new ArrayList<>();
 
     static {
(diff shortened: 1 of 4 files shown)


app/src/main/java/app/revanced/tumblr/patches/TimelineFilterPatch.java
```diff
@@ -0,0 +1,28 @@
...
+        Iterator<TimelineObject<? extends Timelineable>> iterator = timelineObjects.iterator();
+        while (iterator.hasNext()) {
+            TimelineObject<? extends Timelineable> timelineElement = iterator.next();
+            if (timelineElement == null) continue;
+            String elementType = timelineElement.getData().getTimelineObjectType().toString();
+
+            // This dummy gets removed by the TimelineFilterPatch and in its place,
+            // equivalent instructions with a different constant string
+            // will be inserted for each Timeline object type filter.
+            if ("BLOCKED_OBJECT_DUMMY".equals(elementType)) iterator.remove();
```
oSumAtrIX, line 25: How about allocating an array as a field, then letting the patch change it to add new strings, and then instead of this line, you'd loop over the array items?

leumasme: See review in the linked issue for more discussion about this - I have added a warning comment to the method for now, indicating that refactoring it will require a patch update as well.

leumasme:
This does not work with your latest commit in this pr, see https://github.com/ReVanced/revanced-patches/pull/3018#pullrequestreview-1651827823

PR #490, opened 2023-10-02
feat(YouTube - Return YouTube Dislike): Support version `18.37.36`

oSumAtrIX:
Integrations for https://github.com/ReVanced/revanced-patches/pull/3061

### DIFF
--- a/app/src/main/java/app/revanced/integrations/patches/ReturnYouTubeDislikePatch.java
+++ b/app/src/main/java/app/revanced/integrations/patches/ReturnYouTubeDislikePatch.java
@@ -144,23 +144,27 @@ public class ReturnYouTubeDislikePatch {
                                                  @NonNull AtomicReference<CharSequence> textRef,
                                                  @NonNull CharSequence original) {
         try {
-            if (!SettingsEnum.RYD_ENABLED.getBoolean() || PlayerType.getCurrent().isNoneOrHidden()) {
+            if (!SettingsEnum.RYD_ENABLED.getBoolean()) {
                 return original;
             }
 
             String conversionContextString = conversionContext.toString();
             LogHelper.printDebug(() -> "conversionContext: " + conversionContextString);
 
-            final boolean isSegmentedButton;
+            final Spanned replacement;
             if (conversionContextString.contains("|segmented_like_dislike_button.eml|")) {
-                isSegmentedButton = true;
-            } else if (conversionContextString.contains("|dislike_button.eml|")) {
-                isSegmentedButton = false;
+                replacement = ReturnYouTubeDislike.getDislikesSpanForRegularVideo((Spannable) original, true);
+            } else if (conversionContextString.contains("|dislike_button.eml|") ) {
+                // This code path is basically dead because it's only used when spoofing between 17.09.xx and 17.30.xx
+                // but spoofing to that range gives a broken UI layout.
+                // Keep this check here anyways just in case the old litho layout is somehow still used.
+                replacement = ReturnYouTubeDislike.getDislikesSpanForRegularVideo((Spannable) original, false);
+            } else if (conversionContextString.contains("|shorts_dislike_button.eml|")) {
+                replacement = ReturnYouTubeDislike.getDislikeSpanForShort((Spannable) original);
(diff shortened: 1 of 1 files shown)


oSumAtrIX:
@LisoUseInAIKyrios I am not entirely sure of the strategy that is used in the integrations regarding synchronization and caching of video id's, but this PR should be an initial attempt on fixing support of RYD for this version.

oSumAtrIX:
Another bug I have noticed is when you vote, it refreshes the voting.

oSumAtrIX:
The same happens when you exit the app, and enter it again. I can see the pattern here. I believe the issue can be fixed by checking the current playing video id using the old hook.

oSumAtrIX:
So logically, it should be possible to figure out if the current video playing is the current video or the next. This can be done the following way.

1. User swipes down
2. The `newVideoLoaded` hook is called with two videoId's on Shorts
3. Remember the last video IDs that were hooked. 
4. If the user now exits the app, and opens it again, YouTube rebuilds the components.
5. Check if the last remembered video ID is still the same as the last one, the  `newVideoLoaded` hooked.
6. If so, this means, the same Short is playing -> Use the first of the two video id's that was hooked
7. If `newVideoLoaded` is called again, this means the user has swiped down again.

While this works for Shorts, I don't know how well this strategy works in addition to regular videos. The current player type can be checked for that reason.

oSumAtrIX:
I have noticed that the video ID is present directly in the currently rendered video inside the protobuf strings. It is being logged right after the player response video id hook as well as before the hook to change the dislike span.

oSumAtrIX:
@LisoUseInAIKyrios I have pushed some proof of concept changes that completely fix the issue as we are now able to identify the current dislike component directly to the current video id.

oSumAtrIX:
> On first launch (or after clearing the app data), the shorts player is missing the like/dislike/share buttons. They only appear after force closing and relaunching the app. Maybe patching a slightly older or newer version of YouTube might fix this as it seems like a YouTube bug (a ReVanced hide patch probably won't act up only on the first start)

Yes, I have noticed this too. This happens because YouTube does not fully initialize at the very first launch. It is likely one of the filters which causes it, not a tremendous problem though, considering YouTube is involved in this problem.

> Or even better, detect when the user turns off spoofing and manually clear all app files except the preferences.

Yes, I agree, this sounds reasonable. This can be solved in a separate PR. I am not entirely sure how we can manually clear all app files except the preferences, though, a temporal quick solution would be simply letting the user know about this side effect and the solution like you mentioned.

> The only 100% fix is to do the dislikes asynchronously (find the text views for the shorts dislike, or figure out how to force a litho update, or maybe block the player response parsing until RYD has fetched).

This is certainly possible, but would require some more reverse engineering effort to understand how the component framework of Litho works. We have some implementations that work closely with it such as the patch that opens the old quality popup menu. Another starting point would be to see how a click event handler modifies the value.

Another issue I have seen before your commit was that voting caused the dislikes to glitch out, this may be fixed by your commit.

oSumAtrIX:
@LisoUseInAIKyrios Primarily because the release is big, many people would update for the sake of it but miss RYD, so shortly after they would need to update again. This would be annoying, though nothing else prevents us from doing that.

app/src/main/java/app/revanced/integrations/patches/components/ReturnYouTubeDislikeFilterPatch.java
```diff
@@ -0,0 +1,76 @@
...
+
+@RequiresApi(api = Build.VERSION_CODES.N)
+public final class ReturnYouTubeDislikeFilterPatch extends Filter {
+
+    private final ByteArrayFilterGroupList videoIdFilterGroup = new ByteArrayFilterGroupList();
+
+    public ReturnYouTubeDislikeFilterPatch() {
+        pathFilterGroupList.addAll(
+                new StringFilterGroup(SettingsEnum.RYD_SHORTS, "|shorts_dislike_button.eml|")
+        );
```
oSumAtrIX, line 21: Where is this filter used? @LisoUseInAIKyrios

PR #512, opened 2023-11-06
fix(YouTube): Prevent side effects from performing clicks manually

oSumAtrIX:
Fixes https://github.com/ReVanced/revanced-patches/issues/3234.


oSumAtrIX:
I agree, there are a couple places where we have to fix expression consistency in patch names as well as descriptions

PR #537, opened 2023-12-10
fix(YouTube - Announcements): Don't show error toast if there is no internet connection

johnconner122:


app/src/main/java/app/revanced/integrations/patches/announcements/AnnouncementsPatch.java
```diff
@@ -30,6 +30,9 @@ private AnnouncementsPatch() {
 
     @RequiresApi(api = Build.VERSION_CODES.O)
     public static void showAnnouncement(final Activity context) {
+        // Check if there is internet connection
+        if (!ReVancedUtils.isNetworkConnected()) return;
```
oSumAtrIX, line 34: Move this below the setting check.

oSumAtrIX (approved):

PR #539, opened 2023-12-12
refactor(YouTube - Litho filter): Simplify filtering callbacks

LisoUseInAIKyrios:
Minor refactoring of:

- Use methods to add identifier/paths Filter callbacks (less confusing to understand, and simplifies GroupList implementation)
- Flattened `ByteArrayAsStringFilterGroup` into a `ByteArrayFilterGroup` constructor (documentation is easier to understand)
- Added comments of how to filter the protobuffer

### DIFF
--- a/app/src/main/java/app/revanced/integrations/patches/components/LithoFilterPatch.java
+++ b/app/src/main/java/app/revanced/integrations/patches/components/LithoFilterPatch.java
@@ -219,12 +219,14 @@ class ByteArrayFilterGroup extends FilterGroup<byte[]> {
         int matchedLength = 0;
         int matchedIndex = -1;
         if (isEnabled()) {
-            if (failurePatterns == null) {
+            int[][] failures = failurePatterns;
+            if (failures == null) {
                 buildFailurePatterns(); // Lazy load.
+                failures = failurePatterns;
             }
             for (int i = 0, length = filters.length; i < length; i++) {
                 byte[] filter = filters[i];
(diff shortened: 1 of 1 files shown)


app/src/main/java/app/revanced/integrations/patches/components/LithoFilterPatch.java
```diff
@@ -322,42 +310,90 @@ protected ByteTrieSearch createSearchGraph() {
...
-     * @param matchedIndex Matched index of string/array.
-     * @return True if the litho item should be filtered out.
+     * @param contentType  The type of content matched.
+     * @param contentIndex Matched index of the identifier or path.
+     * @return True if the litho component should be filtered out.
      */
-    @SuppressWarnings("rawtypes")
     boolean isFiltered(@Nullable String identifier, String path, byte[] protobufBufferArray,
-                       FilterGroupList matchedList, FilterGroup matchedGroup, int matchedIndex) {
+                       StringFilterGroup matchedGroup, FilterContentType contentType, int contentIndex) {
```
oSumAtrIX, line 379: To reduce boilerplate, it might be a good idea to create a FilterResult. This could hold state on the `contentType` the `matchedGroup`, `contentIntex` and similar so the signature of the method `isFiltered` is reduced. What do you think?

oSumAtrIX (approved):

oSumAtrIX:
What is the purpose of callbacks? What is a sample scenario they are helpful in and how are they used?

oSumAtrIX (approved):

PR #541, opened 2023-12-14
build: bump Gradle dependencies

naxitoo:


oSumAtrIX:
I am not sure, but this should be handled by Dependabot, if not please let me know.

PR #563, opened 2024-02-06
fix(YouTube - Litho Filter): Do not show toast if protobuffer is empty or null

LisoUseInAIKyrios:
Partial fix for https://github.com/ReVanced/revanced-patches/issues/2694

This fixed the error toast, and fixes filtering of non protobuffer layout components.

Protobuffer elements still suffer from inconsistent hiding when using the most recent versions of YT.


oSumAtrIX (commented):
The check method is being overriden by filters, how is that handled? My initial idea was to return in the injection point of the filters hook, but that would prevent filtering anything when the buffer is null as we'd return there

oSumAtrIX (approved):

PR #565, opened 2024-02-07
fix(YouTube - Litho Filter): Ignore null buffers

LisoUseInAIKyrios:
After looking at the patched code (which I should have done to begin with), it's now obvious the buffer can be null when calling into integrations.

So it's possible the buffer is getting cleared just before the call to `LithoFilter`. 
The safe solution is to ignore the null call and use whatever the buffer was last set to.

This may be the actual fix needed to completely resolve the action/flyout items inconsistently hiding in [revanced-patches#2694](https://www.github.com/ReVanced/revanced-patches/issues/2694)


oSumAtrIX:
Unfortunately, I cannot test the changes as I do not experience the issue this PR intends to close. We should contact affected people and do a user test with, for example, @cyberboh @MarcaDian @mad-tunes. If you can, please contact me at osumatrix@revanced.app.

oSumAtrIX (approved):
The changes make sense though.

mad-tunes:
Patches `4.2.0` look good with YT `19.04.37` to me, no toast messages and action buttons hide/show as I set. Thanks

PR #578, opened 2024-03-02
fix(YouTube - Disable suggested video end screen): Disable by default to fix autoplay issues

oSumAtrIX:
Autoplay is only triggered if the end screen is shown. If it is dismissed, which happens if the patch is enabled, autoplay will be stopped.


ILoveOpenSourceApplications:
I thought I saw [@YT-Advanced](https://github.com/YT-Advanced) sharing this somewhere but I can't seem to find it anywhere. This commit  https://github.com/YT-Advanced/ReX-patches/commit/207234a93e7957a03273a93bd5ae07c109d4e53b seems to fix this issue. I'm not sure if it works 100% as I personally don't use autoplay, but during my brief testing it looks like it's working. Just wanted to share this to see if it helps in anyway.

oSumAtrIX:
I have looked into it already but didn't like the implementation and overhead just for this sake. Disabling the patch by default fixes the issue as those who use Autoplay won't mind the end screen video

PR #579, opened 2024-03-02
fix(YouTube - Announcements): Only compare ID to not show same announcement after a fix-up

oSumAtrIX:
Previously, the entire announcement was hashed and compared with the locally saved hash. Sometimes announcements are fixed up, and that alters the hash, so instead, just compare the ID.


oSumAtrIX:
Please open an issue

PR #580, opened 2024-03-03
fix(YouTube - Hide seekbar): Use original seekbar color if Theme patch is not included

LisoUseInAIKyrios:
Fixes https://github.com/ReVanced/revanced-patches/issues/2810

Also noticed hide seekbar timestamp requires a restart to apply.


oSumAtrIX (approved):

oSumAtrIX:
This may fix coloring to red, but it does not fix the linked issue in the sense of the patch being included despite not chosing the coloring patch

oSumAtrIX:
A solution to this would be to make a third patch that handles the hook, then the hide seekbar in feed as well as the color patch can use it independently to each other, for now this solution should be sufficient

oSumAtrIX:
Oh, that works then

PR #582, opened 2024-03-04
fix(Spoof Player Parameter): Playback issues appears when opening a YouTube link

YT-Advanced:
#### Issues: https://github.com/ReVanced/revanced-patches/issues/2819

### DIFF
--- a/app/src/main/java/app/revanced/integrations/youtube/patches/spoof/SpoofSignaturePatch.java
+++ b/app/src/main/java/app/revanced/integrations/youtube/patches/spoof/SpoofSignaturePatch.java
@@ -90,7 +90,7 @@ public class SpoofSignaturePatch {
         try {
             Logger.printDebug(() -> "Original protobuf parameter value: " + parameters);
 
-            if (!Settings.SPOOF_SIGNATURE.get()) {
+            if (parameters == null || !Settings.SPOOF_SIGNATURE.get()) {
                 return parameters;
             }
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/app/revanced/integrations/youtube/patches/spoof/SpoofSignaturePatch.java
```diff
@@ -90,15 +90,15 @@ public static String spoofParameter(String parameters, boolean isShortAndOpening
...
+            if (parameters == null || !Settings.SPOOF_SIGNATURE.get()) {
                 return parameters;
             }
 
             // Clip's player parameters contain a lot of information (e.g. video start and end time or whether it loops)
             // For this reason, the player parameters of a clip are usually very long (150~300 characters).
             // Clips are 60 seconds or less in length, so no spoofing.
             //noinspection AssignmentUsedAsCondition
-            if (useOriginalStoryboardRenderer = parameters.length() > 150 || containsAny(parameters, CLIPS_PARAMETERS)) {
+            if (useOriginalStoryboardRenderer = parameters.length() > 150 || parameters.startsWith(CLIPS_PARAMETERS)) {
```
oSumAtrIX, line 101: containsAny also matches to whatever starts with matches, but why is starts with being used here? Does the clips parameter appear at a later point in the substring? If so why and does just returning the original parameters if the clips parameter is at the beginning of the string only have any other effect?

oSumAtrIX (changes requested):

app/src/main/java/app/revanced/integrations/youtube/patches/spoof/SpoofSignaturePatch.java
YT-Advanced: Yeah right. `KAIB` is also be used when you play video from a link, you can check back my picture that I sent in issues

PR #583, opened 2024-03-05
fix(YouTube - Downloads): Use new task context

LisoUseInAIKyrios:
Integration changes for https://github.com/ReVanced/revanced-patches/pull/2841

### DIFF
--- a/app/src/main/java/app/revanced/integrations/youtube/patches/DownloadsPatch.java
+++ b/app/src/main/java/app/revanced/integrations/youtube/patches/DownloadsPatch.java
@@ -33,6 +33,9 @@ public final class DownloadsPatch {
      *
      * Call if download playlist is pressed, or if download button is used
      * for old spoofed version (both playlists and the player action button).
+     *
+     * Downloading playlists is not supported yet,
+     * as the hooked code does not easily expose the playlist id.
      */
     public static boolean inAppDownloadPlaylistLegacyOnClick(@Nullable String videoId) {
         if (videoId == null || videoId.isEmpty()) {


app/src/main/java/app/revanced/integrations/shared/Utils.java
```diff
@@ -233,6 +233,11 @@ public static Context getContext() {
         return context;
     }
 
+    public static void setContext(Context appContext) {
+        context = appContext;
+        Logger.printDebug(() -> "Set context: " + appContext); // Cannot log before context is set.
```
oSumAtrIX, line 238: Is that the case? The method does not rely on context: https://github.com/revanced/revanced-integrations/blob/bedcef89a5df3846e6613fa216a8cb79143bbedf/app/src/main/java/app/revanced/integrations/shared/Logger.java#L60-L75

app/src/main/java/app/revanced/integrations/youtube/settings/Settings.java
```diff
@@ -335,6 +337,14 @@ public class Settings extends BaseSettings {
         // and more time should be given for users who rarely upgrade.
         migrateOldSettingToNew(DEPRECATED_SB_UUID_OLD_MIGRATION_SETTING, SB_PRIVATE_USER_ID);
 
+
+        // Old spoof versions that no longer work reliably
```
oSumAtrIX, line 341:
```suggestion
        // Old spoof versions that no longer work reliably.
```

app/src/main/java/app/revanced/integrations/youtube/settings/preference/ReVancedPreferenceFragment.java
```diff
@@ -12,14 +19,20 @@
...
     @Override
     protected void initialize() {
         super.initialize();
 
-        // If the preference was included, then initialize it based on the available playback speed
-        Preference defaultSpeedPreference = findPreference(Settings.PLAYBACK_SPEED_DEFAULT.key);
-        if (defaultSpeedPreference instanceof ListPreference) {
-            CustomPlaybackSpeedPatch.initializeListPreference((ListPreference) defaultSpeedPreference);
+        try {
+            // If the preference was included, then initialize it based on the available playback speed
```
oSumAtrIX, line 29:
```suggestion
            // If the preference was included, then initialize it based on the available playback speed.
```

PR #584, opened 2024-03-08
feat(YouTube - Hide layout components): Filter home/search results by keywords

LisoUseInAIKyrios:
Integration changes for https://github.com/ReVanced/revanced-patches/pull/2853

### DIFF
--- a/app/src/main/java/app/revanced/integrations/youtube/patches/NavigationButtonsPatch.java
+++ b/app/src/main/java/app/revanced/integrations/youtube/patches/NavigationButtonsPatch.java
@@ -1,19 +1,22 @@
 package app.revanced.integrations.youtube.patches;
 
-import android.view.View;
-import app.revanced.integrations.youtube.settings.Settings;
-import app.revanced.integrations.youtube.shared.NavigationBar;
+import static app.revanced.integrations.youtube.shared.NavigationBar.NavigationButton;
 
-import java.util.HashMap;
+import android.view.View;
+
+import java.util.EnumMap;
 import java.util.Map;
+
+import app.revanced.integrations.youtube.settings.Settings;
 
 @SuppressWarnings("unused")
 public final class NavigationButtonsPatch {
-    private static final Map<NavigationBar.NavigationButton, Boolean> shouldHideMap = new HashMap<>() {
+
+    private static final Map<NavigationButton, Boolean> shouldHideMap = new EnumMap<>(NavigationButton.class) {
         {
-            put(NavigationBar.NavigationButton.HOME, Settings.HIDE_HOME_BUTTON.get());
-            put(NavigationBar.NavigationButton.CREATE, Settings.HIDE_CREATE_BUTTON.get());
-            put(NavigationBar.NavigationButton.SHORTS, Settings.HIDE_SHORTS_BUTTON.get());
+            put(NavigationButton.HOME, Settings.HIDE_HOME_BUTTON.get());
+            put(NavigationButton.CREATE, Settings.HIDE_CREATE_BUTTON.get());
+            put(NavigationButton.SHORTS, Settings.HIDE_SHORTS_BUTTON.get());
         }
     };
 
@@ -30,7 +33,7 @@ public final class NavigationButtonsPatch {
     /**
      * Injection point.
      */
-    public static void navigationTabCreated(NavigationBar.NavigationButton button, View tabView) {
+    public static void navigationTabCreated(NavigationButton button, View tabView) {
         if (Boolean.TRUE.equals(shouldHideMap.get(button))) {
             tabView.setVisibility(View.GONE);
         }


app/src/main/java/app/revanced/integrations/youtube/TrieSearch.java
```diff
@@ -72,23 +69,14 @@ boolean matches(TrieNode<T> enclosingNode, // Used only for the get character me
 
     static abstract class TrieNode<T> {
         /**
-         * Dummy value used for root node. Value can be anything as it's never referenced.
+         * Dummy value used for root node. Value can be anything as it's never referenced..
```
oSumAtrIX, line 72:
```suggestion
         * Dummy value used for root node. Value can be anything as it's never referenced.
```

app/src/main/java/app/revanced/integrations/youtube/patches/components/HideKeywordContentFilter.java
```diff
@@ -0,0 +1,137 @@
...
+
+        for (String phrase : split) {
+            // Remove any trailing white space the user may have accidentally included.
+            phrase = phrase.stripTrailing();
+            if (phrase.isBlank()) continue;
+
+            if (phrase.length() < MINIMUM_KEYWORD_LENGTH) {
+                Utils.showToastLong(str("revanced_hide_keyword_toast_invalid_length", MINIMUM_KEYWORD_LENGTH, phrase));
+                Utils.showToastLong(str("revanced_hide_keyword_toast_reset"));
+                setting.resetToDefault();
```
oSumAtrIX, line 73: Some people may have many filters and accidentally resetting them may be annoying. Instead, how about setting the setting to the phrases that fulfill the condition? This may apply to custom filters as well.

app/src/main/java/app/revanced/integrations/youtube/patches/components/HideKeywordContentFilter.java
```diff
@@ -0,0 +1,137 @@
...
+            return;
+        }
+
+        // Add common casing that might appear.
+        // If a desired keyword has mixed casing (such as "Matt Whatever")
+        // the user will need to add that exact string.
+        // This could be simplified by adding case insensitive search to the prefix search,
+        // but that also brings a small performance hit.
+        List<String> modifiedKeywords = new ArrayList<>();
+        for (String keyword : keywords) {
```
oSumAtrIX, line 94: Are these hardcoded behaviors of the content filter known to the user? In the description or somewhere, quirky behaviours such as the minimum limit of 3 characters, the casing as well as the locations in which the filter is active and other kinds of behaviors must be known to the user.

app/src/main/java/app/revanced/integrations/youtube/patches/components/HideKeywordContentFilter.java
```diff
@@ -0,0 +1,137 @@
...
+            }
+            // Include first letter capitalization variant.
+            String sentenceCapital = Character.toUpperCase(keyword.charAt(0))
+                    + keyword.substring(1);
+            if (!sentenceCapital.equals(keyword) && !sentenceCapital.equals(upperCase)) {
+                modifiedKeywords.add(sentenceCapital);
+            }
+        }
+        Logger.printDebug(() -> "Using keywords: " + modifiedKeywords);
+        bufferSearch.addPatterns(convertStringsToBytes(modifiedKeywords.toArray(new String[0])));
```
oSumAtrIX, line 114: Duplicates may be filtered here. The user may try enter casings that may have been added by the hardcoded implementation already.

oSumAtrIX (changes requested):

app/src/main/java/app/revanced/integrations/youtube/patches/components/KeywordContentFilter.java
```diff
@@ -0,0 +1,156 @@
...
+import app.revanced.integrations.youtube.settings.Settings;
+
+/**
+ * <pre>
+ * Allows hiding home feed and search results based on keywords and/or channel names.
+ *
+ * Limitations:
+ * - Searching for a keyword phrase will give no search results
+ * - Filtering a channel name can still show Shorts from that channel in the search results
+ * - Some layout component residue will remain, such as the video chapter previews for some search results.
```
oSumAtrIX, line 26: While it is clear that limitations exist, it is not clear why they exist. Adding an explanation for the limitations would solve this.

app/src/main/java/app/revanced/integrations/youtube/patches/components/HideKeywordContentFilter.java
oSumAtrIX: > The 3 character minimum is not explained in the settings, but they will find out when they enter a 1 or 2 character string (it shows an error toast on app startup).

The problem with this is, that it will nuke their filters and that may be annoying as at the point they see their error, their filters would be gone and they'd need to redo them.

app/src/main/java/app/revanced/integrations/youtube/patches/components/KeywordContentFilter.java
```diff
@@ -0,0 +1,156 @@
...
+     */
+    private static final int MINIMUM_KEYWORD_LENGTH = 3;
+
+    /**
+     * Substrings that are always first in the path.
+     */
+    private final StringFilterGroup startsWithFilter = new StringFilterGroup(
+            Settings.HIDE_KEYWORD_CONTENT,
+            "home_video_with_content.eml",
+            "search_video_with_context.eml",
```
oSumAtrIX, line 45: compact_channel.eml could be added to allow filtering for the channel in search results.

app/src/main/java/app/revanced/integrations/youtube/patches/components/KeywordContentFilter.java
```diff
@@ -0,0 +1,156 @@
...
+            "related_video_with_content",
```
oSumAtrIX, line 46: post_base_wrapper.eml could be added to allow filtering in community posts.

app/src/main/java/app/revanced/integrations/youtube/patches/components/KeywordContentFilter.java
```diff
@@ -0,0 +1,156 @@
...
+            "compact_video.eml",
```
oSumAtrIX, line 47: grid_video_wrapper.em appears in channels which features a list of videos. The filter can be used to block keywords in the title.

app/src/main/java/app/revanced/integrations/youtube/patches/components/KeywordContentFilter.java
```diff
@@ -0,0 +1,156 @@
...
+
+    /**
+     * Minimum keyword/phrase length to prevent excessively broad content filtering.
+     */
+    private static final int MINIMUM_KEYWORD_LENGTH = 3;
+
+    /**
+     * Substrings that are always first in the path.
+     */
+    private final StringFilterGroup startsWithFilter = new StringFilterGroup(
```
oSumAtrIX, line 42: Exposing this filter group to the user in an input preference may be a good idea. By default the current values can be added, but in some cases users may want to remove or add their own filters.

app/src/main/java/app/revanced/integrations/youtube/patches/components/KeywordContentFilter.java
```diff
@@ -0,0 +1,156 @@
...
+            Settings.HIDE_KEYWORD_CONTENT,
+            "home_video_with_content.eml",
+            "search_video_with_context.eml",
+            "related_video_with_content",
+            "compact_video.eml",
+            "inline_shorts",
```
oSumAtrIX, line 48: compact_playlist.eml seems to be present under a channels "Playlist" tab. I think it could also appear under other feeds:

image

oSumAtrIX: compact_playlist are the playlist components visible in the screenshot I attached:

image

app/src/main/java/app/revanced/integrations/youtube/patches/components/KeywordContentFilter.java
oSumAtrIX: > Do users want to hide videos for a channel they intentionally navigated to?

In some scenarios, this can be useful. For example, they talk about many subjects if you want to filter for profanity or a specific topic, such as in the LinusTechTips channel.

In other scenarios, that may not be wanted, such as when you want to filter a specific channel entirely, such as Mr Beast, but occasionally want to manually check their channel.

That could be solved by allowing the user to choose the filters via an additional input preference. Alternatively, a couple of switches, such as a "Filter channel videos" switch, could be added for such cases.

app/src/main/java/app/revanced/integrations/youtube/patches/components/KeywordContentFilter.java
oSumAtrIX: > But Shorts use the same layout paths for both home and search, which makes it difficult to selectively turn on/off like this.

I think edge cases like this can be left unconsidered, but the most likely scenarios such as the one you mentioned [here](https://github.com/ReVanced/revanced-integrations/pull/584#discussion_r1519512103) could be turned into a toggleable switch.

app/src/main/java/app/revanced/integrations/youtube/patches/components/KeywordContentFilter.java
oSumAtrIX: Following up on the idea of adding switches, this could also be one, depending on the user, they can filter channel results as well.

app/src/main/java/app/revanced/integrations/youtube/patches/components/KeywordContentFilter.java
oSumAtrIX: There are "Mix" playlists in feeds, but I am not sure if those are the same components. Unfortunately, I can not get them to appear on my feeds, but I can imagine that "compact_playlist" could appear in the feeds, just like "compact_video"

oSumAtrIX:
What do you need the search bar for?

app/src/main/java/app/revanced/integrations/shared/Utils.java
```diff
@@ -196,18 +196,29 @@ public static float getResourceDimension(@NonNull String resourceIdentifierName)
...
+    }
+
     /**
+     * @param searchRecursively If children ViewGroups should also be
+     *                          recursively searched using depth first search.
      * @return The first child view that matches the filter.
      */
     @Nullable
-    public static <T extends View> T getChildView(@NonNull ViewGroup viewGroup, @NonNull MatchFilter filter) {
+    public static <T extends View> T getChildView(@NonNull ViewGroup viewGroup, boolean searchRecursively,
```
oSumAtrIX, line 209: Instead of this new parameter, a separate method can be added that only searches the children.

oSumAtrIX (approved):

PR #591, opened 2024-03-25
chore: Remove unnecessary consumer parameter

oSumAtrIX:
The query parameter was introduced with the intention of being used in the future. It turns out that it is not necessary and can be removed therefor.

### DIFF
--- a/app/src/main/java/app/revanced/integrations/youtube/settings/Settings.java
+++ b/app/src/main/java/app/revanced/integrations/youtube/settings/Settings.java
@@ -345,6 +345,10 @@ public class Settings extends BaseSettings {
         }
 
 
+        // Remove any previously saved announcement consumer (a random generated string).
+        Setting.preferences.saveString("revanced_announcement_consumer", null);
+
+
         // endregion
     }
 }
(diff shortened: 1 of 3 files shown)


app/src/main/java/app/revanced/integrations/youtube/settings/Settings.java
```diff
@@ -196,7 +196,6 @@ public class Settings extends BaseSettings {
     public static final BooleanSetting SPOOF_DEVICE_DIMENSIONS = new BooleanSetting("revanced_spoof_device_dimensions", FALSE, true);
     public static final BooleanSetting BYPASS_URL_REDIRECTS = new BooleanSetting("revanced_bypass_url_redirects", TRUE);
     public static final BooleanSetting ANNOUNCEMENTS = new BooleanSetting("revanced_announcements", TRUE);
-    public static final StringSetting ANNOUNCEMENT_CONSUMER = new StringSetting("revanced_announcement_consumer", "", false, false);
```
oSumAtrIX, line 199: Perhaps there is a way to "delete" the shared preference, instead of just ignoring it like this?

PR #599, opened 2024-03-29
feat(YouTube - GmsCore): Require ignoring battery optimizations

oSumAtrIX:


app/src/main/java/app/revanced/integrations/shared/GmsCoreSupport.java
```diff
@@ -48,40 +49,78 @@ private static void open(String queryOrLink, String message) {
...
-            context.getPackageManager().getPackageInfo(GMS_CORE_PACKAGE_NAME, PackageManager.GET_ACTIVITIES);
+            // Verify, GmsCore is installed.
+            PackageManager manager = activity.getPackageManager();
+            manager.getPackageInfo(GMS_CORE_PACKAGE_NAME, PackageManager.GET_ACTIVITIES);
         } catch (PackageManager.NameNotFoundException exception) {
-            Logger.printInfo(() -> "GmsCore was not found", exception);
-            open(getGmsCoreDownload(), str("gms_core_not_installed_warning"));
-            return;
+            Logger.printDebug(() -> "GmsCore was not found");
+            Utils.showToastLong(str("gms_core_not_installed_warning"));
```
oSumAtrIX, line 59: Whats the reason for not showing a toast when GmsCore is not installed?

app/src/main/java/app/revanced/integrations/shared/GmsCoreSupport.java
```diff
@@ -48,40 +49,78 @@ private static void open(String queryOrLink, String message) {
...
 
-            Logger.printInfo(() -> "GmsCore is not running in the background");
-            open(DONT_KILL_MY_APP_LINK, str("gms_core_not_running_warning"));
+            // Check, if GmsCore is running in the background.
+            try (var client = context.getContentResolver().acquireContentProviderClient(GMS_CORE_PROVIDER)) {
+                if (client == null) {
+                    Logger.printDebug(() -> "GmsCore is not running in the background");
+                    showDoNotKillMyAppDialog(context, "gms_core_not_whitelisted_not_allowed_in_background_message");
+                }
+            }
```
oSumAtrIX, line 105: I kept this check because I am assuming that regardless of fulfilling the check above regarding battery optimizations, additional changes may be necessary. If this is not the case, this check can be removed because the check above would make this one obsolete.

app/src/main/java/app/revanced/integrations/shared/GmsCoreSupport.java
oSumAtrIX: Sorry, I meant dialog.

app/src/main/java/app/revanced/integrations/shared/GmsCoreSupport.java
```diff
@@ -71,12 +71,14 @@ private static void showDoNotKillMyAppDialog(Context context, String messageKey)
...
         Utils.runOnMainThreadDelayed(() -> {
             new AlertDialog.Builder(context)
+                    .setIcon(android.R.drawable.ic_dialog_alert)
                     .setTitle(str("gms_core_not_whitelisted_title"))
                     .setMessage(str(messageKey))
                     .setPositiveButton(android.R.string.ok, (dialog, id) -> {
                         open(DONT_KILL_MY_APP_LINK);
                         System.exit(0);
                     })
+                    .setCancelable(false)
```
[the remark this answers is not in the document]
oSumAtrIX: If we intend the user to dismiss it, we should add a close/continue anyways/later button. The user might be in a hurry for example where this would be better than asking the user to change some settings.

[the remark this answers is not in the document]
oSumAtrIX: A proper affordance instead of a hidden, seemingly unintended bypass is acceptable because the next time they launch the app, the dialogue will nag them again. The only issue I can think of is if the user skips it and experiences problems while using the app, then proceeds to report an issue. Users allegedly experience playback issues while GmsCore is killed in the background, although I hardly believe that as that would crash the app.

app/src/main/java/app/revanced/integrations/youtube/patches/announcements/AnnouncementsPatch.java
```diff
@@ -35,6 +37,12 @@ public static void showAnnouncement(final Activity context) {
         // Check if there is internet connection
         if (!Utils.isNetworkConnected()) return;
 
+        // If GmsCore is not installed, the https call will always fail with a security exception.
+        if (GmsCoreSupport.gmsIsNotInstalled()) {
```
oSumAtrIX, line 41: This makes the patch dependent on GmsCore support patch. Whats the reason for security exceptions?

app/src/main/java/app/revanced/integrations/shared/GmsCoreSupport.java
```diff
@@ -40,48 +50,98 @@ private static void open(String queryOrLink, String message) {
...
-            open(getGmsCoreDownload(), str("gms_core_not_installed_warning"));
-            return;
-        }
+            // Verify GmsCore is installed.
+            try {
+                PackageManager manager = context.getPackageManager();
+                manager.getPackageInfo(GMS_CORE_PACKAGE_NAME, PackageManager.GET_ACTIVITIES);
+            } catch (PackageManager.NameNotFoundException exception) {
+                Logger.printDebug(() -> "GmsCore was not found");
+                gmsIsNotInstalled = true;
```
oSumAtrIX, line 98: If GmsCore fails to run in the background, should this also be true?

app/src/main/java/app/revanced/integrations/shared/GmsCoreSupport.java
```diff
@@ -40,48 +50,99 @@ private static void open(String queryOrLink, String message) {
         intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
         Utils.getContext().startActivity(intent);
 
-        // Gracefully exit, otherwise without Gms the app crashes and Android can nag the user.
+        // Gracefully exit, otherwise the broken app will continue to run.
         System.exit(0);
     }
 
+    private static void showToastOrDialog(Context context, String toastMessageKey, String dialogMessageKey, String link) {
```
oSumAtrIX, line 57: Unfortunately, on WSA it seems the dialog is also not shown, and the app crashes if GmsCore is not installed. This might be due to GMS not being installed at all although GmsCore support patch changes most references to GmsCore. 

I have captured the following logs:

This one occurs from a check in the app. I have managed to circumvent it once by returning the correct enum integer value in one of the check methods (0), but the app may still crash:
```
03-30 03:23:56.809  1999  2144 E GooglePlayServicesUtil: GooglePlayServices not available due to error 9
03-30 03:23:56.809  1999  2144 E ttq     : Attempted to use SSL unpatched. Google Play Services unavailable.
03-30 03:23:56.809  1999  2144 E ttq     : ord
03-30 03:23:56.809  1999  2144 E ttq     :      at orf.d(PG:6)
```
```
03-30 03:19:36.874  1282  1389 E YouTube : [GservicesConfigHelper][preloadGservicesCacheBlocking] java.lang.SecurityException: Failed to find provider com.google.android.gsf.gservices for user 0; expected to find a valid ContentProvider for this authority
03-30 03:19:36.954  1282  1339 E ActivityThread: Failed to find provider info for com.google.android.gsf.gservices
03-30 03:19:37.070  1282  1452 E DynamiteModule: Invalid GmsCore APK, remote loading disabled.
03-30 03:19:37.071  1282  1452 E DynamiteModule: Invalid GmsCore APK, remote loading disabled.
03-30 03:19:37.073  1282  1459 E FirebaseInstanceId: Google Play services missing or without correct permission.
0
```

oSumAtrIX:
Yes, the toast is shown just like pre PR, although the spamming issue is back the same way. It was an offtopic issue so I merged anyways

PR #601, opened 2024-03-31
fix(YouTube - GmsCore support): Prompt to disable battery optimizations, if not done already

LisoUseInAIKyrios:
Integrations for https://github.com/ReVanced/revanced-patches/pull/2958


app/src/main/java/app/revanced/integrations/shared/GmsCoreSupport.java
```diff
@@ -81,40 +89,73 @@ public static void checkGmsCore(Context context) {
...
             Logger.printException(() -> "checkGmsCore failure", ex);
         }
     }
 
+    private static void checkIfUserAllowedTurningOffBatteryOptimizations(Context context) {
+        // Check if the user actually turned off power optimizations for GmsCore.
+        // This can be done by patching the Activity and overriding the
+        // onActivityResult() method, but that only works for YT
+        // and not YT Music because Music is not hooked to an Activity.
+        // Instead, check back in 20 seconds and show a toast if it's still not enabled.
```
oSumAtrIX, line 133: Probably this should be worded generalized. If at all, YouTube and YouTube Music should be mentioned as example cases, since the patch is generalized for any app.

app/src/main/java/app/revanced/integrations/shared/GmsCoreSupport.java
```diff
@@ -81,40 +89,73 @@ public static void checkGmsCore(Context context) {
...
+                        "gms_core_dialog_continue_text",
+                        (dialog, id) -> openGmsDisableBatteryOptimizationsIntent((Activity) context),
+                        dialog -> checkIfUserAllowedTurningOffBatteryOptimizations(context));
+            }
         } catch (Exception ex) {
             Logger.printException(() -> "checkGmsCore failure", ex);
         }
     }
 
+    private static void checkIfUserAllowedTurningOffBatteryOptimizations(Context context) {
```
oSumAtrIX, line 128: On dismiss, should the user really be nagged to disable battery optimizations?

oSumAtrIX (commented):
In patches, "turn off" was suggested to change to "disable", the same should be done here if we use "disable" in patches.

oSumAtrIX (approved):

PR #615, opened 2024-04-12
feat(YouTube - Hide Shorts components): Hide tagged products, hide search suggestions

LisoUseInAIKyrios:
Integration changes for https://github.com/ReVanced/revanced-patches/pull/3019

### DIFF
--- a/app/src/main/java/app/revanced/integrations/youtube/patches/components/ShortsFilter.java
+++ b/app/src/main/java/app/revanced/integrations/youtube/patches/components/ShortsFilter.java
@@ -67,8 +67,8 @@ public final class ShortsFilter extends Filter {
         shortsCompactFeedVideoBuffer = new ByteArrayFilterGroup(null, "/frame0.jpg");
 
         // Shorts player components.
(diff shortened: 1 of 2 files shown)


app/src/main/java/app/revanced/integrations/youtube/patches/components/ShortsFilter.java
```diff
@@ -130,62 +123,60 @@ public ShortsFilter() {
...
         //
-        suggestedActions.addAll(
+        suggestedActionsGroupList.addAll(
                 new ByteArrayFilterGroup(
                         Settings.HIDE_SHORTS_SHOP_BUTTON,
                         "yt_outline_bag_"
                 ),
+                new ByteArrayFilterGroup(
+                        Settings.HIDE_SHORTS_TAGGED_PRODUCTS,
+                        "PAproduct_listZ"
```
oSumAtrIX, line 178: Is there an image of the component's appearance or where it appears? If this is the same component as the other buttons, such as Shop (grey transparent button), maybe it's better and more straightforward for the user to either toggle the button on or off instead of having so many options for every possible kind the button can be.

PR #622, opened 2024-04-23
feat(YouTube - Hide ads): Add option to hide the 'Visit store' button on channel pages

oSumAtrIX:


app/src/main/java/app/revanced/integrations/youtube/patches/components/AdsFilter.java
```diff
@@ -100,6 +104,16 @@ public AdsFilter() {
...
         );
 
+        channelProfile = new StringFilterGroup(
+                null,
+                "channel_profile.eml"
+        );
+
+        visitStoreButton = new ByteArrayFilterGroup(
+                Settings.HIDE_VISIT_STORE_BUTTON,
+                "header_store_button"
```
oSumAtrIX, line 114: @LisoUseInAIKyrios It may be enough to just filter for this without an additional check for channelProfile. Should we get rid of channelProfile?

PR #631, opened 2024-05-11
feat(Boost For Reddit): Add `Fix /s/ links` patch

OctoNezd:
Multi-part PR:
- Part 1: https://github.com/ReVanced/revanced-patches/pull/3154

This PR fixes https://github.com/ReVanced/revanced-patches/issues/478

Code is essentially copy-paste from patch for Sync, but with more exception handling and adding the Authorization header to prevent reddit from giving the client page "log in or provide token or else" without proper Location: header.


OctoNezd:
I really hate the database query part of this - however I couldn't find anything inside context that includes access token or any sort of HTTP client - so its the best solution I managed to come up with :(

My network is censored weirdly - I cant open reddit without VPN, and reddit blocks VPNs

app/src/main/java/app/revanced/integrations/boostforreddit/FixSLinksPatch.java
```diff
@@ -0,0 +1,93 @@
...
+            return null;
+        }
+        tokenCursor.moveToFirst();
+        String token = tokenCursor.getString(0);
+        tokenCursor.close();
+        redditDb.close();
+        return token;
+    }
+
+    public static String resolveSLink(Context context, String link) {
```
oSumAtrIX, line 41: Probably a good idea to introduce an interface class and use singletons to access the implementing methods via a patch since there's now two and possibly in the future, more resolve s link patches

app/src/main/java/app/revanced/integrations/boostforreddit/FixSLinksPatch.java
```diff
@@ -0,0 +1,93 @@
...
+import app.revanced.integrations.shared.Logger;
+
+import java.io.File;
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.net.SocketTimeoutException;
+import java.net.URL;
+
+public final class FixSLinksPatch {
+    public static String getUserAccessToken(Context context) {
```
oSumAtrIX, line 20: Should be private

app/src/main/java/app/revanced/integrations/boostforreddit/FixSLinksPatch.java
```diff
@@ -0,0 +1,93 @@
...
+
+                connection.connect();
+                String location = connection.getHeaderField("location");
+                connection.disconnect();
+
+                // Restore the original strict mode policy.
+                StrictMode.setThreadPolicy(currentPolicy);
+
+                Logger.printInfo(() -> "Resolved " + link + " -> " + location);
+                // For some reason using requireNotNull or similar ends up in java.lang.ExceptionInInitializerError,
```
oSumAtrIX, line 61: Probably a throwable is raised instead of an exception

oSumAtrIX:
> I really hate the database query part of this - however I couldn't find anything inside context that includes access token or any sort of HTTP client - so its the best solution I managed to come up with :(
> 
> My network is censored weirdly - I cant open reddit without VPN, and reddit blocks VPNs

This isn't related to s links in any way? In that case they should be not part of this PR, since we squash the PR into a single commit

app/src/main/java/app/revanced/integrations/boostforreddit/FixSLinksPatch.java
OctoNezd: I am not particularly versed in java - but perhaps abstract class would make more sense? Considering that both patches right now do basically the same thing, only `getUserAccessToken` needs to be overridden.

app/src/main/java/app/revanced/integrations/boostforreddit/FixSLinksPatch.java
```diff
@@ -0,0 +1,93 @@
...
+
+        return link;
+    }
+
+    @NonNull
+    private static HttpURLConnection getHttpURLConnection(Context context, String link) throws IOException {
+        URL url = new URL(link);
+        HttpURLConnection connection = (HttpURLConnection) url.openConnection();
+        connection.setRequestProperty("User-Agent", "android:app.revanced.reddit:v1.0 (by /u/spez)");
+        // Auth token is needed if users IP is blacklisted by reddit, e.g. VPN
```
oSumAtrIX, line 83: Any reason why this isn't necessary in the Sync integrations?

app/src/main/java/app/revanced/integrations/boostforreddit/FixSLinksPatch.java
```diff
@@ -0,0 +1,93 @@
...
+        String accessToken = getUserAccessToken(context);
```
oSumAtrIX, line 84: This should probably be cached and not called Everytime

oSumAtrIX (changes requested):

OctoNezd:
This is related to s-links - to get slink reliably (incl. banned networks) Authorization header with user token must be supplied, else reddit gives 403 without proper Location.

app/src/main/java/app/revanced/integrations/boostforreddit/FixSLinksPatch.java
oSumAtrIX: Checks out to me 👍

app/src/main/java/app/revanced/integrations/boostforreddit/FixSLinksPatch.java
OctoNezd: This is necessary in Sync as well if you have IP address that is banned by reddit, as far as I understand it - gonna check later. Original PR probably didnt count for that because it was introduced before reddit started banning VPN IPs left and right.

app/src/main/java/app/revanced/integrations/boostforreddit/FixSLinksPatch.java
OctoNezd: Error still gets logged by exception handler however

app/src/main/java/app/revanced/integrations/boostforreddit/FixSLinksPatch.java
OctoNezd: You mean the getUserAccessToken?

app/src/main/java/app/revanced/integrations/boostforreddit/FixSLinksPatch.java
OctoNezd: Yep - just checked, Sync gets null instead of proper URL on VPN that was blacklisted by reddit:
```05-11 21:46:21.754  2452  2452 I revanced: FixSLinksPatch: Resolving https://www.reddit.com/r/PSP/s/JoTORdHSwS
05-11 21:46:22.556  2452  2452 I revanced: FixSLinksPatch: Resolved https://www.reddit.com/r/PSP/s/JoTORdHSwS -> null```

File: app/src/main/java/app/revanced/integrations/boostforreddit/FixSLinksPatch.java (hunk 5 above)
Comment (OctoNezd) [reply app/src/main/java/app/revanced/integrations/boostforreddit/FixSLinksPatch.java:84]: In any case, do you have any suggestions as to where to store the cached value?

File: app/src/main/java/app/revanced/integrations/boostforreddit/FixSLinksPatch.java (hunk 2 above)
Comment (OctoNezd) [reply app/src/main/java/app/revanced/integrations/boostforreddit/FixSLinksPatch.java]: Should it still be private? Interfaces don't allow private+default

**Comment by OctoNezd** [comment]
Do I add Sync for reddit switch to interface in this PR or should I make a separate one after that one gets merged?

**Comment by OctoNezd** [comment]
On note of implementing support for Sync for reddit:

Sync doesn't store access tokens in its DB (unlike boost) - only refresh tokens. The access token in sync seem to be kept inside com.laurencedawson.reddit_sync.singleton.AccountSingleton - but I have no idea how one can even access it from integration. Unless the patch would also patch the AccountSingleton method that sets the Authorization header and somehow transfer between these two.

**Comment by OctoNezd** [comment]
Tested on Boost 1.12.12 and Sync v23.06.30-13:39. 

Resolution no longer halts the UI.

Everything hopefully should be good, gonna run around with boost for a few days and check if everything gonna be okay.

**Review by oSumAtrIX — CHANGES_REQUESTED** [review]
The abstraction can be improved. Theres still some duplicate implementations such as openInAppBrowser or resolveSLink. Instead of copying the implementation, try to reduce duplicates. In the example of openInAppBrowser you could for example add a field of type Activity and then set the field in the constructor of the instance with either activity for Sync and Boost, then the base class implementation could reference it. You would then not need to duplicate the implementation like you do now in the implementing classes.

File: app/src/main/java/app/revanced/integrations/shared/fixes/slink/BaseFixSLinksPatch.java [hunk 6]
```diff
@@ -0,0 +1,116 @@
...
+            Logger.printInfo(() -> "Resolving " + link);
+            if (link.endsWith("#bypass")) {
+                openInAppBrowser(context, link);
+                return ResolveResult.DO_NOTHING;
+            }
+            String accessToken = getUserAccessToken(context);
+            if (accessToken == null) {
+                // this is terrible!
+                // but we need to get access_token to properly auth request, especially if user is
+                // using VPN or similar
```
oSumAtrIX, line 36: Comments start with a big letter and end with punctuation. Try to be neutral.

OctoNezd:
> In the example of openInAppBrowser you could for example add a field of type Activity and then set the field in the constructor of the instance with either activity for Sync and Boost, then the base class implementation could reference it

Are you suggesting that both Boost & Sync use the same class, and inside constructor there is some sort of switch-case that determines webview activity based on package name?

app/src/main/java/app/revanced/integrations/boostforreddit/FixSLinksPatch.java
```diff
@@ -0,0 +1,51 @@
...
+            case DO_NOTHING:
+                ret = true;
+                break;
+            default:
+                break;
+        }
+        return ret;
+    }
+
+    public static void getAccessToken(String access_token) {
```
OctoNezd, line 48: I think this name can create confusion with getUserAccessToken, considering it sets access token that will be used. Perhaps something more of appGetAccessToken?

oSumAtrIX (commented):
Left some refactoring changes and TODOs in the code. Please check on them and test the refactoring changes (same for the patches part). Apart from that lgtm.

app/src/main/java/app/revanced/integrations/shared/fixes/slink/BaseFixSLinksPatch.java
```diff
@@ -0,0 +1,116 @@
...
+import app.revanced.integrations.shared.Logger;
+import app.revanced.integrations.shared.Utils;
+
+import static app.revanced.integrations.shared.Utils.getContext;
+
+public abstract class BaseFixSLinksPatch {
+    String accessToken = null;
+    public String pendingUrl = null;
+
+    public ResolveResult performResolution(Context context, String link) {
```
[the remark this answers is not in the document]
OctoNezd: Removed it and replaced with getcontext

OctoNezd:
I am not sure on todo you added:

```
                    // TODO: Check if this ends in java.lang.ExceptionInInitializerError.
                    Objects.requireNonNull(location, "Location is null");
```
Wouldnt that be caught by catch (Exception e)?

oSumAtrIX:
> Wouldnt that be caught by catch (Exception e)?

You originally left a comment claiming that Objects.requireNonNull was causing a RuntimeException, but given that you were not able to reproduce again the question is redundant.

oSumAtrIX (approved):

PR #635, opened 2024-05-18
feat(YouTube - Navigation buttons): Add option to hide navigation button labels

MarcaDian:
integrations for https://github.com/ReVanced/revanced-patches/pull/3189


oSumAtrIX:
I think solving with a patch time rather than a runtime solution is simpler and more consistent.

oSumAtrIX (approved):

PR #638, opened 2024-05-22
fix(YouTube - Client spoof): Spoof iOS client model to fix various side effects

oSumAtrIX:
This PR also changes the default client spoof to iOS, as the only side effect known now is that HDR is not working.

### DIFF
--- a/app/src/main/java/app/revanced/integrations/youtube/patches/spoof/SpoofClientPatch.java
+++ b/app/src/main/java/app/revanced/integrations/youtube/patches/spoof/SpoofClientPatch.java
@@ -4,6 +4,7 @@ import static app.revanced.integrations.youtube.patches.spoof.requests.Storyboar
 
 import android.net.Uri;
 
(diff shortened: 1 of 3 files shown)


app/src/main/java/app/revanced/integrations/youtube/patches/spoof/SpoofClientPatch.java
```diff
@@ -250,14 +262,16 @@ public static int getRecommendedLevel(int originalLevel) {
     }
 
     private enum ClientType {
-        ANDROID_TESTSUITE(30, "1.9"),
-        IOS(5, Utils.getAppVersionName());
+        ANDROID_TESTSUITE(30, Build.MODEL, "1.9"),
+        IOS(5, "iPhone15,4", Utils.getAppVersionName());
```
YT-Advanced, line 266: Uhm I think we should use Iphone 15 Pro Max: `iPhone16,2`

PR #644, opened 2024-05-28
fix(YouTube - Spoof client): Disable AV1 if not supported by spoofing an older iOS device

LisoUseInAIKyrios:
Change iOS device to iPhone XS Max, to (maybe/probably) reduce the chance of using AV1 streams.

### DIFF
--- a/app/src/main/java/app/revanced/integrations/youtube/patches/spoof/SpoofClientPatch.java
+++ b/app/src/main/java/app/revanced/integrations/youtube/patches/spoof/SpoofClientPatch.java
@@ -112,14 +112,14 @@ public class SpoofClientPatch {
         // https://dumps.tadiphone.dev/dumps/oculus/monterey/-/blob/vr_monterey-user-7.1.1-NGI77B-256550.6810.0-release-keys/system/system/build.prop
         // version 1.37 is not the latest, but it works with livestream audio only playback.
         ANDROID_VR(28, "Quest", "1.37"),
-        // 15,3 = iPhone 14 Pro Max.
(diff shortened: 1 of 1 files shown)


oSumAtrIX:
Don't devices that support AV1 profit of these streams?

oSumAtrIX (approved):

oSumAtrIX:
@sodawithoutsparkles

SodaWithoutSparkles:
My phone has a Dimensity 8100 which supports AV1 decode according to the manufacturer's site. This could maybe solve some of the issues. 

I've also identified a few other possible causes for the stuttering. I'll now use the stats for nerds feature on every video now to see if I could identify some trends.

I can test this next morning with CLI, I still need to test the other issue at patches/3238

P.S. I wonder if the `deviceHasAV1HardwareDecoding` func passes if the device only supports decode not encode. I might be wrong here, but it seems that it just checks if it has either decode or encode?

SodaWithoutSparkles:
I was still getting avc1. Now starting ADB to check the logs

SodaWithoutSparkles:
From logcat I can see that it did detect that my phone supports AV1 hardware decode: 

```shell
05-29 17:45:05.475 22902 23020 D revanced: SpoofClientPatch: Device supports AV1 hardware decoding.
```

Nevertheless it is still using format 299 (avc1 1920x1080@60) when using [this video](https://youtu.be/njX2bu-_Vw4) as a test. It is not only limited to this video tho, all videos are still avc1.

SodaWithoutSparkles:
It works as intended then

PR #668, opened 2024-07-24
fix(YouTube - Client Spoof): Unlock more qualities by spoofing the iOS client user agent 

zainarbani:
Integrations for: [ReVanced/revanced-patches#3468](https://github.com/ReVanced/revanced-patches/pull/3468)

### DIFF
--- a/app/src/main/java/app/revanced/integrations/youtube/patches/spoof/SpoofClientPatch.java
+++ b/app/src/main/java/app/revanced/integrations/youtube/patches/spoof/SpoofClientPatch.java
@@ -103,6 +103,33 @@ public class SpoofClientPatch {
     /**
      * Injection point.
      */
+    public static String getOsVersion(String originalOsVersion) {
+        if (SPOOF_CLIENT_ENABLED) {
+            return SPOOF_CLIENT_TYPE.osversion;
+        }
+
+        return originalOsVersion;
+    }
+
+    /**
+     * Injection point.
+     */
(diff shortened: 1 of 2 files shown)


0xrxL:
@zainarbani Some useful links to update client version:

[iOS](https://apps.apple.com/us/app/youtube-watch-listen-stream/id544007664) - Last: 19.29.1

[VR](https://www.meta.com/experiences/2002317119880945/) - Last: 1.58.14

OxrxL:
> The VR version acts up when using older/newer versions (particularly with live streams).
> 
> Unless changing the version is shown to fix limitations, it's better to keep using the current versions.

Same thing for iOS?

zainarbani:
@0xrxL Can you try this patch and see wether HDR uses AV01 or VP9.2? (you can use my forks build)

0xrxL:
> @0xrxL Can you try this patch and see wether HDR uses AV01 or VP9.2? (you can use my forks build)

Only AV1 unlock HDR:

Screenshot_20240725_085813_YouTube

Screenshot_20240725_085752_YouTube

oSumAtrIX:
Does iOS play HDR vp9 streams

0xrxL:
> Does iOS play HDR vp9 streams

If it's a question: nope, at least in my case. iOS spoof can just play HDR videos with av01 codec.

zainarbani:
> Does iOS play HDR vp9 streams

It play vp9 HDR starting from iOS 14

oSumAtrIX (approved):

PR #673, opened 2024-08-04
fix(YouTube - Spoof client): Restore livestream audio only playback with iOS spoofing

LisoUseInAIKyrios:
Integration changes for https://github.com/ReVanced/revanced-patches/pull/3504


oSumAtrIX:
It should probably be reverted. DNS blocking is something you need to fix on your end.

oSumAtrIX (approved):

PR #675, opened 2024-08-13
feat(YouTube): Add `Check watch history domain name resolution` patch

LisoUseInAIKyrios:
Integration changes for https://github.com/ReVanced/revanced-patches/pull/3537

### DIFF
--- a/app/src/main/java/app/revanced/integrations/youtube/patches/CheckWatchHistoryDomainNameResolutionPatch.java
+++ b/app/src/main/java/app/revanced/integrations/youtube/patches/CheckWatchHistoryDomainNameResolutionPatch.java
@@ -46,7 +46,7 @@ public class CheckWatchHistoryDomainNameResolutionPatch {
      * Checks if s.youtube.com is blacklisted and playback history will fail to work.
      */
     public static void checkDnsResolver(Activity context) {
-        if (!Utils.isNetworkConnected() || !Settings.CHECK_WATCH_HISTORY_DNS_RESOLVER.get()) return;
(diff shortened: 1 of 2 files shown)


app/src/main/java/app/revanced/integrations/youtube/patches/DnsResolverVerificationPatch.java
```diff
@@ -0,0 +1,76 @@
...
+import java.net.UnknownHostException;
+
+import app.revanced.integrations.shared.Logger;
+import app.revanced.integrations.shared.Utils;
+import app.revanced.integrations.youtube.settings.Settings;
+
+@SuppressWarnings("unused")
+public class DnsResolverVerificationPatch {
+
+    private static final String YOUTUBE_PLAYBACK_HISTORY_API_ADDRESS = "s.youtube.com";
```
oSumAtrIX, line 18: `HISTORY_TRACKING_ENDPOINT` is simpler

app/src/main/java/app/revanced/integrations/youtube/patches/DnsResolverVerificationPatch.java
```diff
@@ -0,0 +1,76 @@
...
+
+    private static final String SINKHOLE_IPV4 = "0.0.0.0";
```
oSumAtrIX, line 20: Some DNS blockers may return 127.0.0.1 or any other loopback IP address. Calling the endpoint with HEAD to read a HTTP response code should be more failsafe & doesn't require checking for IPv6 either.

PR #681, opened 2024-08-28
feat(YouTube - Keyword filter): Add syntax to match whole keywords and not substrings

LisoUseInAIKyrios:
Integration changes for https://github.com/ReVanced/revanced-patches/pull/3592


app/src/main/java/app/revanced/integrations/youtube/patches/components/KeywordContentFilter.java
```diff
@@ -216,23 +220,167 @@ private static String capitalizeAllFirstLetters(String sentence) {
...
             }
+
+            i += Character.charCount(codePoint);
+        }
+
+        return false;
+    }
+
+    /**
+     * @return If the phrase will will hide all videos. Not an exhaustive check.
```
oSumAtrIX, line 253:
```suggestion
     * @return If the phrase will hide all videos. Not an exhaustive check.
```

PR #682, opened 2024-08-28
fix(YouTube - Hide layout components): Hide search result shelf header

ILoveOpenSourceApplications:
Closes [#8](https://github.com/ReVanced/revanced-patches/issues/8)

### DIFF
--- a/app/src/main/java/app/revanced/integrations/youtube/patches/components/LayoutComponentsFilter.java
+++ b/app/src/main/java/app/revanced/integrations/youtube/patches/components/LayoutComponentsFilter.java
@@ -198,7 +198,7 @@ public final class LayoutComponentsFilter extends Filter {
 
         searchResultShelfHeader = new StringFilterGroup(
                 Settings.HIDE_SEARCH_RESULT_SHELF_HEADER,
-                "shelf_header.eml"
+                "compact_channel.eml"
         );
 
         notifyMe = new StringFilterGroup(


oSumAtrIX (changes requested):
Compact channel hides channels in some places like search the last time I saw it. I think it also only happens in tablet mode

ILoveOpenSourceApplications:
This is how the header looks currently:
Screenshot_20240828_233918_YouTube

ILoveOpenSourceApplications:
They have merged the header with the channel profile, basically making into one. If needed then the name of the patch can be changed accordingly.

ILoveOpenSourceApplications:
Maybe something like `Hide channel profile in search results`?

ILoveOpenSourceApplications:
A more user-centric approach would be to present users with a choice while also allowing users to determine their preference which would enhance the overall user experience.

oSumAtrIX:
With PR https://github.com/ReVanced/revanced-patches/pull/3462, ReVanced Integrations has been merged into ReVanced Patches. This PR must be reopened in the [ReVanced Patches](https://github.com/ReVanced/revanced-patches) repository and adjusted to the new build system. ReVanced Integration is now understood as "extensions" according to the ReVanced Patcher API. As extensions are now on a per-patch basis, this PR may be subject to a separate extension. If not, the PR can change the shared extension in [ReVanced Patches](https://github.com/ReVanced/revanced-patches).

Thanks, for the PR.

PR #689, opened 2024-09-07
feat(YouTube): Support version `~19.31`

zainarbani:


0xrxL:
@LisoUseInAIKyrios Keep attention to Cairo Navigation Bar style. It cause a player downwards misalignment, during miniplayer switching gesture.

app/src/main/java/app/revanced/integrations/youtube/patches/MiniplayerPatch.java
```diff
@@ -52,6 +58,40 @@ public boolean isModern() {
...
+
+            // Instead of resetting, clamp the size at the bounds.
+            dipWidth = Math.max(WIDTH_DIP_MIN, Math.min(dipWidth, WIDTH_DIP_MAX));
+            Settings.MINIPLAYER_WIDTH_DIP.save(dipWidth);
+        }
+
+        MINIPLAYER_SIZE = dipWidth;
+    }
+
+    private static final boolean IS_19_20_OR_GREATER = Utils.getAppVersionName().compareTo("19.20.00") >= 0;
```
oSumAtrIX, line 90: How do we properly maintain fragments of code like this in the future? Say we update a patch and drop an old version. We would have to return here and many other places in the code to change lines like these. This is not optimal. I would prefer the patch only to support the latest version if it means getting rid of fragmented code like this that burdens maintenance and code quality in the future.

app/src/main/java/app/revanced/integrations/youtube/shared/VersionCheck.java
```diff
@@ -0,0 +1,10 @@
+package app.revanced.integrations.youtube.shared;
+
+import app.revanced.integrations.shared.Utils;
+
+public class VersionCheck {
```
oSumAtrIX, line 5: Should this be called `VersionCheckPatch` for consistency like the other patches?

PR #693, opened 2024-09-18
fix(YouTube - Check environment): Only use fields available since Android 8

oSumAtrIX:


oSumAtrIX:
image

Noticed this is not being used. @LisoUseInAIKyrios is this intentional?

oSumAtrIX:
Sorry I meant the unused field `PATCH_SERIAL`.

PR #709, opened 2024-10-03
fix(YouTube - Spoof video streams): Handle app left open for a long time

LisoUseInAIKyrios:
Seems to fix random playback errors when the app is left open for a long period of time.

### DIFF
--- a/app/src/main/java/app/revanced/integrations/youtube/patches/spoof/SpoofVideoStreamsPatch.java
+++ b/app/src/main/java/app/revanced/integrations/youtube/patches/spoof/SpoofVideoStreamsPatch.java
@@ -69,15 +69,9 @@ public class SpoofVideoStreamsPatch {
                 String path = originalUri.getPath();
 
                 if (path != null && path.contains("initplayback")) {
-                    // This previously returned an unreachable URL and it seemed to work for nearly all situations,
-                    // but an unreachable url would cause playback failure if the app changes to/from mobile
-                    // or if the app is left open for a long period of time.
-                    //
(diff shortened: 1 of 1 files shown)


app/src/main/java/app/revanced/integrations/youtube/patches/spoof/SpoofVideoStreamsPatch.java
```diff
@@ -69,9 +69,9 @@ public static String blockInitPlaybackRequest(String originalUrlString) {
                 String path = originalUri.getPath();
 
                 if (path != null && path.contains("initplayback")) {
-                    Logger.printDebug(() -> "Blocking 'initplayback' by returning unreachable url");
+                    Logger.printDebug(() -> "Blocking 'initplayback' by clearing query");
```
oSumAtrIX, line 72: This would still request YouTube. Why not return an unreachable URL?

oSumAtrIX (approved):
LGTM apart from the one review

app/src/main/java/app/revanced/integrations/youtube/patches/spoof/SpoofVideoStreamsPatch.java
oSumAtrIX: What fixes it is using the original request headers. They may expire, so they have to be refreshed; this PR handles it by using the original request headers, so they are always up to date. Blocking the initplayback request by clearing the query should not have any effect on this. An unreachable URL should be used to avoid bothering YouTube with error requests.

PR #710, opened 2024-10-04
feat(Sync for Reddit): Add `Fix video downloads` patch

1fexd:
Re-implements MPD manifest interpretation to fix video track not being saved - see https://github.com/ReVanced/revanced-patches/pull/3739


app/src/main/java/app/revanced/integrations/syncforreddit/internal/RedditVideoPlaylistParser.kt
```diff
@@ -0,0 +1,51 @@
+package app.revanced.integrations.syncforreddit.internal
+
+import org.w3c.dom.Element
+import java.io.ByteArrayInputStream
+import javax.xml.parsers.DocumentBuilderFactory
+
+object RedditVideoPlaylistParser {
```
oSumAtrIX, line 7: You can move this to the Java class. No need to use Kotlin, because Kotlin ships its stdlib and when the app you patch has a different version that can cause conflicts.

oSumAtrIX (commented):
Why is a reimplementation of the parser needed? Can the existing parser in Sync not be fixed for that?

1fexd:
> Why is a reimplementation of the parser needed? Can the existing parser in Sync not be fixed for that?

Just rewriting it as an integrations patch was less tedious and allowed me to easily test it standalone

oSumAtrIX (approved):

PR #727, opened 2024-10-24
fix(YouTube - Settings): Use multiline preference title for localized languages

LisoUseInAIKyrios:
Fixes preference titles clipping when using localized langauges.

Many of the localized preference titles do not fit into 1 line and many are unreadable.

This PR sets the titles to multiline if the device is using a localized language.

The English strings do not require multiline, and the 2 or 3 English preferences that overrun are only partially clipped and still readable.  Making those couple of preferences multiline looks weird and out of place, and visually it is better to clip the text and keep all titles 1 line.

<details>
<summary>Before (Single line)</summary>

Screenshot_20241024_083249_YouTube

</details>

<details>
<summary>After (Multiline)</summary>

Screenshot_20241024_080418_YouTube

</details>

### DIFF
--- a/app/src/main/java/app/revanced/integrations/shared/Utils.java
+++ b/app/src/main/java/app/revanced/integrations/shared/Utils.java
@@ -13,7 +13,6 @@ import android.os.Bundle;
 import android.os.Handler;
 import android.os.Looper;
 import android.preference.Preference;
-import android.preference.PreferenceCategory;
 import android.preference.PreferenceGroup;
 import android.preference.PreferenceScreen;
 import android.view.View;
@@ -746,8 +745,8 @@ public class Utils {
             Preference pref = group.getPreference(i);
             pref.setSingleLineTitle(false);
 
-            if (pref instanceof PreferenceCategory) {
-                setPreferenceTitlesToMultiLineIfNeeded((PreferenceCategory) pref);
+            if (pref instanceof PreferenceGroup) {
+                setPreferenceTitlesToMultiLineIfNeeded((PreferenceGroup) pref);
             }
         }
     }
--- a/app/src/main/java/app/revanced/integrations/shared/Utils.java
+++ b/app/src/main/java/app/revanced/integrations/shared/Utils.java
@@ -13,6 +13,7 @@ import android.os.Bundle;
 import android.os.Handler;
 import android.os.Looper;
 import android.preference.Preference;
+import android.preference.PreferenceCategory;
 import android.preference.PreferenceGroup;
 import android.preference.PreferenceScreen;
 import android.view.View;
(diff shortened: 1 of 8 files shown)


MarcaDian:
This is a very important PR for us, because very often we have to cut the title translation to fit in screen.
However, during testing, I noticed that changes occurred only in Sponsorblock or RYD. In native ReVanced settings, it still gets cropped.
<details>

Screenshot_2024-10-24-19-01-20-402-edit_app revanced android youtube
Screenshot_2024-10-24-19-01-36-576-edit_app revanced android youtube

LisoUseInAIKyrios:
Did you change the language in the YouTube settings?  If so you have to force restart the app because not all the the menus are fully reloaded by YouTube.

MarcaDian:
I switched languages English <-> Ukrainian (and force restart) and nothing changed.


There are also titles clipping in English.
<details>

app/src/main/java/app/revanced/integrations/shared/Utils.java
```diff
@@ -722,6 +723,35 @@ public static void sortPreferenceGroups(@NonNull PreferenceGroup group) {
...
+        }
+
+        String deviceLanguage = Utils.getContext().getResources().getConfiguration().locale.getLanguage();
+        if (deviceLanguage.equals("en")) {
+            return;
+        }
+
+        for (int i = 0, prefCount = group.getPreferenceCount(); i < prefCount; i++) {
+            Preference pref = group.getPreference(i);
+            pref.setSingleLineTitle(false);
```
oSumAtrIX, line 746: Can that not be configured globally for all preferences?

LisoUseInAIKyrios: As far as I know, it cannot.

oSumAtrIX: Seems like the  preference serializer could add `android:singleLineTitle="false"` according to https://stackoverflow.com/questions/19616238/how-to-wrap-a-preference-title-really.

LisoUseInAIKyrios: Yes, except that won't work for RYD or SponsorBlock.  And it can't be selectively applied to only apply to localized languages.

oSumAtrIX: Shouldn't it be applied to any language? Currently, RYD and SB inject their own preference screens without the framework in the patches. So that should probably be unified.

LisoUseInAIKyrios: English doesn't need multiline, and the 2 or 3 places where a title is clipped it looks really weird and out of place if they're multiline.

Visually it looks much better for those few preferences to be clipped by a few letters, than to be outliers with double title lines.



RYD and especially SB have a bunch of logic specific to their settings that would make migrating to xml more complicated than it's worth.

oSumAtrIX: > English doesn't need multiline, and the 2 or 3 places where a title is clipped it looks really weird and out of place if they're multiline. Visually it looks much better for those few preferences to be clipped by a few letters, than to be outliers with double title lines.

Why would English be the exception? English can also be clipped and is subject to be wrapped like any other language if it is too long. The text should only clip if you can expand and read it fully in some way. Otherwise, it just breaks accessibility. It is okay if it looks weird as long as it is readable, and in the rare cases it does, this should be fine.

> RYD and especially SB have a bunch of logic specific to their settings that would make migrating to xml more complicated than it's worth.

Custom preferences would be necessary, although these already exist. For the most part it would just be a matter of refactoring them to be used via XML. I remember some plans to move everything away from XML, but keeping it consistent until then is good.

LisoUseInAIKyrios: I can only find 2 places in the English strings where it clips.  They are still readable:

----
Settings

----

And changing to multiline for these 2 settings looks out of place:
Settings

----
I think visually it's better to clip these two titles, and have a consistent layout in the settings.  The English strings are heavily scrutinized and tweaked, and generally don't have multiline issues.

oSumAtrIX: The wrapping should be agnostic to the strings. We may add new ones in the future or change strings so they get clipped. WIthout making them multiline we have to think about how to format them to avoid that and treat them exceptionally from other languages.

LisoUseInAIKyrios: Taking consideration to make preference titles single line has always been true, so nothing is new here.

We've got around the issue by adding extra text to the summary, which is the norm since preference titles are ideally intended to be 1 line (thats why multiline isn't the default) and summaries are always multiline.

In the future if some large English text cannot fit into a one line title then this behavior can changed then.  But for now I think consistency is better if it means clipping a few letters off two  English preferences.

oSumAtrIX:
With PR https://github.com/ReVanced/revanced-patches/pull/3462, ReVanced Integrations has been merged into ReVanced Patches. This PR must be reopened in the [ReVanced Patches](https://github.com/ReVanced/revanced-patches) repository and adjusted to the new build system. ReVanced Integration is now understood as "extensions" according to the ReVanced Patcher API. As extensions are now on a per-patch basis, this PR may be subject to a separate extension. If not, the PR can change the shared extension in [ReVanced Patches](https://github.com/ReVanced/revanced-patches).

Thanks, for the PR.

